PR #1, opened 2015-10-22
Initial boilerplate

vogdb:
See https://github.com/cargomedia/pulsar-rest-api or https://github.com/cargomedia/socket-redis
- license
- CI
- binary
- etc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: node_js
+node_js:
+  - "0.11"
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Cargo Media
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/package.json
+++ b/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "cm-janus",
+  "description": "",
+  "version": "0.0.1",
+  "author": "Cargomedia",
+  "license": "MIT",
+  "main": "",
+  "bin": {
+    "cm-janus": "./bin/cm-janus"
+  },
+  "keywords": [],
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:cargomedia/cm-janus.git"
+  },
+  "bugs": {
+    "url": "https://github.com/cargomedia/cm-janus/issues"
+  },
+  "scripts": {
+  },
+  "dependencies": {
+  },
+  "devDependencies": {
+  }
+}


vogdb:
@tomaszdurka please review.

tomaszdurka:
Looks ok. You probably could add dummy binary :)

tomaszdurka:
Basic class/file structure:
CmJanus
- Core/Proxy
	- ties everything together
	- maps requests to acl
	- forwards requests if needed
- HttpServer
	- /audio/publish (streamName, user)
	- /video/publish (streamName, user)
- WebsocketServer (?)
- Acl (user, role)
- CMApiClient
	- authenticate(user, role/action)
	- send publish, subscribe, unpublish, unsubscribe information
- JanusGateway.WebsocketClient
	- negotiate protocol, address for video streams
- JanusGateway.HttpClient
	- audio bridge communication

kris-lab:
Few notes:
- websocket server is not needed (for browser clients etc)
- full bidirectional communication cm-janus to janus can be done via websocket client (monitoring, sessions, streams, killing etc)
- httpserver will expose endpoints for browser + cm app. I guess will be on the same port?

njam:
Regarding "CMApiClient: authenticate":
- We don't need authentication for clients. We can just send along the `sessionId` in the data bag to the CM application, which can extract it and do authentication. This way *cm-janus* doesn't need to know much about it.
- For authentication of the CM application itself I think we could just configure an `apiKey` in the config file, and we check if it matches for any HTTP API calls?

kris-lab:
> For authentication of the CM application itself I think we could just configure an apiKey in the config file, and we check if it matches for any HTTP API calls?

Yes, agree. Also there will be NGINX with SSL termination on the public IP. Nginx could only pass requests to upstream which starts `/{audio,video}/{publish,subscribe}`.  Then `cm-janus` will expose `/{status,kill,other}/something` on the same endpoint but will be accessible only in private IP network?

tomaszdurka:
@kris-lab fyi
We had a talk today and we decided that it's actually not needed to put authentiction logic into cm-janus (not even http calls to cm). Authentication will be done solely in cm-application while cm-janus will be in private network only. Clients (browser, satellite, etc) will make HTTP call to cm-application - not cm-janus.

kris-lab:
In general I agree. However since we move a lot of logic to the `CM` instead of `cm-janus` it will be more internal translator instead of bridge! I have to still investigate WebRTC architecture more in depth then! My main concerns (not really investigated) is that:
- browser prepares SDP offer (codec, port, ip, protocol e.g. VP8/5004/UDP/192.23.23.23) where Janus can connect for pushing data (in case of our Streaming/Receive plugin + Audio/Receive plugin)
- janus prepares SDP offer (codec, port, ip, protocol e.g. VP8/5004/UDP/192.23.23.23) where browser WebRTC client can connect (in case of Audio/Send plugin where we send data to janus)

As I said, previously all steps of SDP has been done with JS/Janus plugin in the browser (ICE/STUN etc). Currently we have to handle this workflow of SDPs during translation Browser->CM_APP->CM_Janus->Janus and back!

I will update soon as I know more!

vogdb:
@kris-lab the sooner the better. Please also see #5. It would be nice if we can have a call together.

kris-lab:
Agree. The workflows are a bit complicated. I have initial results:
- audioroom: `SDP/offer` is done by Browser and send to Janus which evaluate `SDP/offer` into `SDP/answer` and send to Browser. Then comes `trickle` (regulation) mechanism - read more below...
- streaming: `SDP/offer` is done by Janus and send to browser which evaluate `SDP/offer` into `SDP/answer` (set local SDP description and remote SDP description to the PeerConnection) and send it back to Janus. It is first step of negotiation. Unfortunately here comes the `trickle` mechanism which allows browser to send `iceCandidates` and updates to SDP offer/answer. The thing is that SDP doesn't have to be full (only media type, name, codecs and SHA fingerprint). In next step the Peers can exchange and even update in realtime the SDP params. It allows to dynamically change IP/Port interface and routing based on bandwidth and performance. All info comes from built-in browse API but must be exchanged by signalling layer (currently WebSocket).

I will still investigate deeper but for now it looks we would have to handle many events!

let's discuss asap

njam:
Researched with Chris:

Illustration/debugging of ICE trickling: https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/

ICE trickling with Janus when using mobile carrier network and a public TURN server:
screen shot 2015-10-26 at 20 26 31

Idea: Take [janus.js](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js), copy it to `cm-janus.js` in this repo and simplify to our needs.
It seems there will be quite a lot of messages from the browser to janus (one for every trickle (`onicecandidate`)). Maybe it's better to let the browser talk to `cm-janus` directly over websockets after all? (and in turn do authentication via RPC calls to CM).

tomaszdurka:
Regarding the architecture decision about moving authenticaton from cm-janus to cm, we were (at least me) not aware about all the difficulties we will encounter. Seeing how complicated negotiation is, how many steps/events it requires it is questionable to go that direction. Also decision we made before about using simple http call instead of websocket for negotation doesn't seem reasonable and rather complicates than simplify things.

I understand that much simpler soluton (and vote for it) is to proxy websocket with RPC call as authenticaton method.

PR #5, opened 2015-10-23
Implement dummy Http server with required endpoints

vogdb:
HttpServer
- /audio/publish (streamName)
- /video/publish (streamName)
- /audio/subscribe (streamName)
- /video/subscribe (streamName)

### DIFF
--- a/lib/cm-rest-server.js
+++ b/lib/cm-rest-server.js
@@ -1,10 +1,14 @@
 var _ = require('underscore');
+var config = require('config');
+var logger = require('./logger');
 var express = require('express');
 var http = require('http');
 var bodyParser = require('body-parser');
+var getLog = logger.getLog;
 var JanusClient = require('./janus-client');
 
 var CmRestServer = function() {
+  logger.configure(config.get('app.logPath').toString());
   var app = express();
   var router = express.Router();
   var server = http.createServer(app);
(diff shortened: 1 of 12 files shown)


lib/cm-rest-server.js
```diff
@@ -0,0 +1,59 @@
+var express = require('express');
+var http = require('http');
+var bodyParser = require('body-parser');
+var janusClient = require('janus-client');
+
+var CmRestServer = function() {
+  var app = express();
+  var router = express.Router();
```
tomaszdurka: Maybe rather pass router as dependency and store on the instance of CmRestServer?
Also let's call it RestServer? I think CM should be reserved for classes which directly interact with CM application, wdyt?

vogdb:
@njam FYI

vogdb:
Currently me and Vladimir are reading about WebRTC architecture cause we've stuck with the problem of setting up of PeerConnection between Janus and Subscriber over cm-janus. The current progress is that we have a Rest server that can return an JSEP+SDP of stream by stream id. How can subscriber instantiate a video channel by that JSEP+SDP is a question at least to me.

cc @krislab @njam @tomaszdurka

vogdb:
@kris-lab please see the comment above.

PR #6, opened 2015-10-27
implement Janus proxy

vogdb:

PR #7, opened 2015-10-28
Define HTTP server

fvovan:
Authentication based on defined secret.

Actions:
- `/stopStream` (using clientKey/special identifier)
- `/status` (list existing streamchannels and their streams) - is it really needed?

### DIFF
--- a/lib/cm-http-server.js
+++ b/lib/cm-http-server.js
@@ -1,0 +1,60 @@
+var express = require('express');
+var http = require('http');
+var bodyParser = require('body-parser');
+var _ = require('underscore');
+
+var CMHttpServer = function() {
+  var app = express();
+  var router = express.Router();
+  var server = http.createServer(app);
+
+  app.use(bodyParser.urlencoded({extended: false}));
+  app.use(bodyParser.json());
+  app.use(router);
+  app.disable('x-powered-by');
+  app.disable('etag');
+
+  this.installRoutes(router);
+
+  //IMPORTANT! Do not remove unused `next`.
+  app.use(function(err, req, res, next) {
+    var errorMessage = err.message || 'Unexpected error';
+    //getLog().error(errorMessage);
+    res.status(err.code || 500).send({error: errorMessage});
+  });
+
+  process.on('uncaughtException', function(err) {
+    //getLog().error('Uncaught exception ' + err.message);
+    process.nextTick(process.exit(1));
+  });
+
+  var listenPort = '8888';
+  server.listen(listenPort);
+  console.log('HTTP server on port ' + listenPort + ' started');
+};
+
+CMHttpServer.prototype.installRoutes = function(router) {
+  var self = this;
+
+  router.use(function(req, res, next) {
+    if (req.body['apiKey'] != '123fish') {
+      res.sendStatus(403);
+    } else {
+      next();
+    }
+  });
+
+  router.post('/stopStream', function subscribe(req, res, next) {
+    var params = _.extend({}, req.params, req.body);
+    var streamId = +params['streamId'];
+    res.send('Stop stream received: ' + streamId);
+  });
+
+  router.get('/status', function subscribe(req, res, next) {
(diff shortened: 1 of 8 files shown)


fvovan:
@vogdb please review.
I had to pull your changes from https://github.com/cargomedia/cm-janus/pull/6 to keep all things here.

lib/cm-http-server.js
```diff
@@ -0,0 +1,60 @@
+var express = require('express');
+var http = require('http');
+var bodyParser = require('body-parser');
+var _ = require('underscore');
+
+var CMHttpServer = function() {
```
tomaszdurka: I wouldn't call it `CM...` server as it's not really CM related.

tomaszdurka:
Can you replace all `console.log()`s with `getLog().debug()`?

vogdb:
Also lets change `getLog().debug()` to `log.debug()`.

tomaszdurka:
Wdyt about:
- Storing `router` and `server` on `HttpServer` as instance variables.
- Adding `listen` method which would start listening for http requests. I think it's not good to listen directly from constructor.

fvovan:
agree about separate method for lister but would call it `start` as did for WS server.
and what's the reason to move `router` to instance variable?

fvovan:
@tomaszdurka please review

vogdb:
So, what notation do we use for our dependencies? `"minilog": "^2.1.0"` or `"minilog": "~2.1.0"`?

vogdb:
Can we optimize logger form this:
```js
var Logger = require('./logger');

var log = new Logger(config.get('app.logPath')).getLog();
```
to this
```js
var log = require('./logger');
```
?

Include somehow log instantiation and configuration into './logger' itself.

fvovan:
>So, what notation do we use for our dependencies? "minilog": "^2.1.0" or "minilog": "~2.1.0"?

It's just default one now. AFAIK they mean different things.

vogdb:
> It's just default one now. AFAIK they mean different things. 

Yes. That is why asked :)

fvovan:
but why not leave default values? Is it bad practice in nodejs?
> Can we optimize logger form this?

Then it will be tightly coupled with Application and its Config.

vogdb:
> Then it will be tightly coupled with Application and its Config. 

I'm not sure that it is bad under this environment.

vogdb:
>> Then it will be tightly coupled with Application and its Config. 

> I'm not sure that it is bad under this environment.

It's just we have to create `Logger` in every file where we want to use logs and that I don't like

tomaszdurka:
@fvovan let's pass logger as dependency injection, wdyt?

fvovan:
@tomaszdurka  is done, but no sure if it is more beautiful now.
probably it will be messy every time when you need for logger to debug something you will have to change constructor signature (plus assigning it to instance variable) , instead of just adding 1 line to beginning of the file. but most likely it is okay

tomaszdurka:
Not sure if prettier, but imo better. It's kinda strange as we anyway create all other dependencies inside constructor. We should rather call it `logger` instead of `log`.

fvovan:
true, my concerns were only for "temporary usage". That's why it's a quite common practice to make a Logger as a singleton or  inject it via traits.

vogdb:
@tomaszdurka why DI for logger? O_o Do you want to pass it to all classes as DI? I think we also need to use it in `Auth` class. And you want to have as DI there? In pulsar projects we never used it as DI. Why now?

tomaszdurka:
Yes, I already see problems with this. Probably it's not so necessary especially in languages like Javascript.

tomaszdurka:
Oh reasons are simple - my point of view changed/improved :) But I agree that it will be messy with passing Logger all way down.

vogdb:
as always

tomaszdurka:
@vogdb I think you are right, I've reverted changes in #9

PR #10, opened 2015-10-29
Introduce RPC client for cm-janus-api

fvovan:
Methods:
- publish (notify cm-app when publish is created)
- unpublish
- subscribe
- unsubscribe


fvovan:
It's very basic now, I used the same signatures as for Wowza
@tomaszdurka could you review it please
and what these methods should return to callers?

bin/cm-janus
```diff
@@ -1,4 +1,6 @@
 #!/bin/env node
+
+process.env['NODE_CONFIG_DIR'] = __dirname + '/config';
```
tomaszdurka: So I don't think this is required as default behaviour is too look for `config` in the root dir. Do you have global `NODE_CONFIG_DIR` on your machine set?

lib/cm-http-client.js
```diff
@@ -0,0 +1,74 @@
+var logger = require('./logger');
+var request = require('request');
+
+/**
+ * @param {String} baseUrl
+ * @param {Number} clientKey
+ */
+CMHttpClient = function(baseUrl, clientKey) {
```
tomaszdurka: Shouldn't this be abstracted from `http` in case of usage. What about `CMApiClient`? Not sure.

lib/cm-http-client.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+/**
+ * @param {String} streamName
+ * @param {Number} start
+ * @param {JSON} data
+ */
+CMHttpClient.prototype.publish = function(streamName, start, data) {
+  request.post(
+    this.baseUrl + '/CM_Janus_RpcEndpoints.publish',
```
tomaszdurka: Probably `baseUrl` should not include `/rpc` prefix.

bin/config/default.json
```diff
@@ -5,5 +5,8 @@
   "httpServer" : {
     "port" : 8888,
     "apiKey" : "123fish"
+  },
+  "cmHttpClient": {
+    "cmUrl" : "http://www.fuckbook.dev.cargomedia.ch/"
```
tomaszdurka: Please don't put fb urls publicly ! :panda_face:

bin/cm-janus
fvovan: Actually it was "cherrypicked" from Alexey's https://github.com/cargomedia/cm-janus/commit/4c458a77fd04c317688cdcebb477d28ab3fb0509
Not sure what is the best way to do it. @vogdb wdyt?

vogdb: It's not looked in the root dir. It's looked in the directory where the binary file is. @tomaszdurka what is the alternative to this? how do you want to set the config location?

tomaszdurka: That's what I meant by root dir (root dir of binary :) ).

I am not sure how this should be done tbh. Seems that's the only choice, I've looked into the code of library and nothing.
Then let's put config out of this directory (`bin` to top level) and set `NODE_CONFIG_DIR` to `path.join(__dirname,  '../config')` ?

Or maybe we should have our own config file `lib/config.js` which would do this? 
Or we should not care about it and pass config dir env evertime we run it?

bin/config/default.json
```diff
@@ -5,5 +5,8 @@
   "httpServer" : {
     "port" : 8888,
     "apiKey" : "123fish"
+  },
+  "cmHttpClient": {
+    "cmUrl" : "http://www.fooboo.dev.cargomedia.ch"
```
tomaszdurka: We usually use `http://www.cm.dev` in those cases.

bin/cm-janus
vogdb: > Then let's put config out of this directory (`bin` to top level) and set `NODE_CONFIG_DIR` to `path.join(__dirname,  '../config')` ?

That is a default behaviour of `config` library. We don't need to set `NODE_CONFIG_DIR` if it is placed here `path.join(__dirname,  '../config')`.

What exactly do you want to achieve here with config files?

fvovan: Nothing special, just general configs with ability to have separate settings per stage and merge by key.

tomaszdurka: Wot, for me default behaviour was it is placed in `__dirname + '/config`... no?

fvovan: I removed it.
@tomaszdurka  please review

vogdb:
@fvovan can you please implement the same Config logic as in pulsar-rest-api, so we can specify a custom filepath for the config file. See the example here https://github.com/cargomedia/pulsar-rest-api/blob/master/bin/pulsar-rest-api#L7-L8. @tomaszdurka please confirm.

tomaszdurka:
@vogdb oh yes we talked with Vladimir about it already. Problably done in another pull-request.

@fvovan btw I think we should use promise-based request here. I will implement.

tomaszdurka:
@vogdb please review :)

vogdb:
is that all?

tomaszdurka:
Yes it is now ! :)

vogdb:
lgtm

PR #11, opened 2015-10-29
Detach of Streaming plugin during the actual streaming kills the proxy

fvovan:
How to model: Open `streamingtest.html` -> Watch any stream -> Click **Stop** button (see the picture).

screen shot 2015-10-29 at 18 08 16

@fvovan @tomaszdurka FYI

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -27,7 +27,7 @@ var HttpServer = function() {
   });
 
   process.on('uncaughtException', function(err) {
-    logger.error('Uncaught exception ' + err.message);
+    logger.error('Uncaught exception ' + err.stack || err.message);
     process.nextTick(process.exit(1));
   });
 };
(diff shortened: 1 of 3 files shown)


fvovan:
@vogdb please review.

vogdb:
It's fine for now but we should fix the root of this. We need to intercept any break in both (janus and browser) connections and shutdown/reconnect appropriately.
@tomaszdurka your thoughts?

fvovan:
we can close Janus side WS connection when receive 'close' on Browser's one.

vogdb:
We are already doing it =) It just does not work because the closing of connection is not immediate =(
Please add me as collaborator to your fork of cm-janus

fvovan:
Ah true, i mixed it with JanusConnection. 
You are added

vogdb:
Thanks! The fix was a one-liner. @c949cc0ef48dcfca41bda68094fb00e5adedad95 shouldn't be needed.

vogdb:
@fvovan please verify

fvovan:
@vogdb , thanks,  it works
Shall I delete WS status checking or leave it?

vogdb:
I guess it's not needed but I like the idea of verifying that connection is open, we can improve the queue mechanism and possibly to queue those `dead letters`.

PR #8, opened 2015-10-29
Define message interpreters

vogdb:
### Browser -> cm-janus
We should deny illegal actions/plugins.
We should instantiate handler per plugin and attach listener to specific `handleId` at that time. We should store it per connection.
- `Streaming`
  - `{"request":"watch"}` holds the stream id. 
- `AudioBridge`
  - ?

### Janus -> cm-janus
In particular, for each handle involving a PeerConnection Janus provides notifications about its current state. To do so, the following events may be received as well:

- `webrtcup`: ICE and DTLS succeeded, and so Janus correctly established a PeerConnection with the user/application;
  - trigger `onSubscribe`, `onPublish` events
- `media`: whether Janus is receiving (receiving: true/false) audio/video (type: "audio/video") on this PeerConnection;
- `hangup`: the PeerConnection was closed, either by Janus or by the user/application, and as such cannot be used anymore.
  - trigger `onUnSubscribe`, `onUnPublish` events

Above events could be intercepted and its data passed to cm applications via rpc calls.

### DIFF
--- a/bin/cm-janus
+++ b/bin/cm-janus
@@ -1,4 +1,6 @@
 #!/bin/env node
+
+process.env['NODE_CONFIG_DIR'] = __dirname + '/config';
 var CMJanus = require('../lib/index.js');
 
 var proxy = new CMJanus.JanusProxy('8188', 'ws://198.23.87.26:8188/janus');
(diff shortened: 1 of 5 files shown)


vogdb:
I'm waiting here for #10.

fvovan:
actually it's done (though all methods are just stubs ) but question about config is still open.

vogdb:
Yes, I understand)

tomaszdurka:
Closing in favor of #12.
@vogdb please reopen if disagree. Thanks.

PR #13, opened 2015-10-31
Make config system a la pulsar-api

fvovan:


fvovan:
@tomaszdurka , @vogdb  could you review it please
It affected logger as we need to pass Config instance downside.
If it's okay I will adjust our WS server to use it.

vogdb:
Very good job! Ideally it would be nice to have the API of the previous module `config` that could be accessed by `require('config').get('propertyName')` + the ability to specify the filepath for it manually through the command line argument.

@tomaszdurka wdyt?

fvovan:
But it has the ability to specify config path. I just moved it to bin/cm-janus file. It's more proper place imo.

vogdb:
You mean that the the previous module `config` had the ability to specify the config path? Cause I saw that it could specify only the config directory and names of files are not subjects to change.

fvovan:
Nono, I meant 'pulsar-style config.
```
var argv = minimist(process.argv.slice(2));
var configPath = argv.c || __dirname + '/config.yaml';
```
But actually  'node-config' loaded filename could be adjusted via `NODE_ENV` env var and used together with `NODE_CONFIG` dir option it can specify full path.

tomaszdurka:
@fvovan please take a look again and merge if ok.

PR #15, opened 2015-11-04
Implement `status` and `stopStream` of http-server

vogdb:


vogdb:
@tomaszdurka please review.

lib/http-server.js
```diff
@@ -51,13 +52,23 @@ HttpServer.prototype.installRoutes = function(router) {
...
+    }
   });
 
   router.get('/status', function subscribe(req, res, next) {
-    var params = _.extend({}, req.params, req.body);
-    res.send('Status received');
+    var result = _.reduce(streams.list, function(memo, stream) {
+      memo[stream.id] = stream.channelName;
+      return memo;
+    }, {});
```
tomaszdurka: I know object should be perfectly fine, but I think array is bit better for this one. Also this could mess up with non-unique ids (which should probably not happen). Wdyt?

vogdb: I think that perfectly designed IDs should be unique as the master who mastered them ;-)

vogdb:
@tomaszdurka please review

lib/proxy-connection.js
```diff
@@ -78,6 +116,36 @@ ProxyConnection.prototype.close = function() {
...
+ * @param {String} streamId
+ */
+ProxyConnection.prototype.stopStream = function(streamId) {
+  this.janusConnection.send({
+    janus: 'message',
+    body: {request: 'stop'},
+    transaction: this._randomTransactionId(),
+    session_id: this.sessionId,
+    handle_id: this._getPluginByStreamId(streamId).id
+  });
```
tomaszdurka: Hmmm, why do you handle all of this and not simply close connection?

lib/transactions.js
```diff
@@ -20,6 +20,17 @@ Transactions.prototype.find = function(id) {
 };
 
 /**
+ * @param {Function} transaction
+ * @param {Object} message
+ */
+Transactions.prototype.execute = function(transaction, message) {
+  if ('ack' != message['janus']) {
```
tomaszdurka: Why not?

lib/proxy-connection.js
```diff
@@ -21,7 +23,7 @@ ProxyConnection.prototype.processMessage = function(message) {
   if (message.transaction) {
     var pendingTransaction = this.transactions.find(message.transaction);
     if (pendingTransaction) {
-      pendingTransaction(message);
+      this.transactions.execute(pendingTransaction, message);
```
tomaszdurka: If you delegate that to transactions you should rather pass transactionId, not the whole transaction?

lib/proxy-connection.js
```diff
@@ -41,6 +49,40 @@ ProxyConnection.prototype.processMessage = function(message) {
...
+  this.transactions.add(message.transaction, function(response) {
+    switch (response.janus) {
+      case 'success':
+        self.sessionId = response.data.id;
+        break;
+      case 'error':
+        //TODO Inform user that create didn't work. Should we close here the connection?
+        break;
+    }
+  });
```
tomaszdurka: Shall you remove transaction here?

lib/streams.js
```diff
@@ -19,12 +19,20 @@ Streams.prototype.remove = function(stream) {
 };
 
 /**
- * @param {ProxyConnection} connection
+ * @param {String} streamId
+ * @returns {Stream|null}
+ */
+Streams.prototype.get = function(streamId) {
```
tomaszdurka: Use `find` as discussed.

lib/transactions.js
tomaszdurka: I don't like it, but let's keep it for simpler callbacks.

lib/transactions.js
```diff
@@ -20,6 +20,17 @@ Transactions.prototype.find = function(id) {
 };
 
 /**
+ * @param {Function} transaction
+ * @param {Object} message
+ */
+Transactions.prototype.execute = function(transaction, message) {
```
tomaszdurka: Let's pass id instead of transaction.

lib/proxy-connection.js
```diff
@@ -78,6 +116,36 @@ ProxyConnection.prototype.close = function() {
...
+ProxyConnection.prototype._getPluginByStreamId = function(streamId) {
+  return _.find(this.plugins, function(plugin) {
+    return streamId == plugin.stream.id;
+  });
+};
+
+/**
+ * @returns {String}
+ */
+ProxyConnection.prototype._randomTransactionId = function() {
```
tomaszdurka: Let's copy janus.js function.
https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L205
or 
`Math.random().toString(36).substring(2, 12)`

lib/proxy-connection.js
```diff
@@ -78,6 +116,36 @@ ProxyConnection.prototype.close = function() {
...
+  });
+};
+
+/**
+ * @param {String} streamId
+ * @returns
+ */
+ProxyConnection.prototype._getPluginByStreamId = function(streamId) {
+  return _.find(this.plugins, function(plugin) {
+    return streamId == plugin.stream.id;
```
tomaszdurka: `plugin.stream` can be undefined in edge cases, then it fails.

lib/proxy-connection.js
```diff
@@ -78,6 +116,36 @@ ProxyConnection.prototype.close = function() {
 };
 
 /**
+ * @param {String} streamId
+ */
+ProxyConnection.prototype.stopStream = function(streamId) {
+  this.janusConnection.send({
```
tomaszdurka: Create transaction, handle possible errors.

lib/http-server.js
```diff
@@ -51,13 +52,22 @@ HttpServer.prototype.installRoutes = function(router) {
 
   router.post('/stopStream', function subscribe(req, res, next) {
     var params = _.extend({}, req.params, req.body);
-    var streamId = +params['streamId'];
-    res.send('Stop stream received: ' + streamId);
+    var streamId = ('' + params['streamId']).trim();
```
tomaszdurka: Use `String()`/`Number()`. Don't trim - expect correct string.

lib/proxy-connection.js
```diff
@@ -21,7 +23,7 @@ ProxyConnection.prototype.processMessage = function(message) {
   if (message.transaction) {
     var pendingTransaction = this.transactions.find(message.transaction);
```
tomaszdurka: Replace `find` with `has` to not allow executing transaction from outside.

lib/http-server.js
tomaszdurka: Also, let's call this `streamKey` instead of `streamId` as we call it in CM like that.

lib/proxy-connection.js
vogdb: Lets keep this ability yet, so we always have a way to handle the exceptional cases.

lib/proxy-connection.js
vogdb: no need

vogdb:
@tomaszdurka please review

lib/http-server.js
```diff
@@ -51,13 +52,23 @@ HttpServer.prototype.installRoutes = function(router) {
 
   router.post('/stopStream', function subscribe(req, res, next) {
     var params = _.extend({}, req.params, req.body);
-    var streamId = +params['streamId'];
-    res.send('Stop stream received: ' + streamId);
+    var streamId = String(params['streamId']);
+    var stream = streams.find(streamId);
+    if (stream) {
+      stream.proxyConnection.stopStream(stream.id);
+      res.send('Stop stream received: ' + streamId);
```
tomaszdurka: Wasn't there a comment about status/succes/error?

lib/proxy-connection.js
```diff
@@ -78,6 +115,43 @@ ProxyConnection.prototype.close = function() {
...
 
 /**
+ * @param {String} streamId
+ */
+ProxyConnection.prototype.stopStream = function(streamId) {
+  var transactionId = this._generateTransactionId();
+  this.transactions.add(transactionId, function(response) {
+    if (response.janus === 'error') {
+      //TODO cmApiClient.???
+    }
```
tomaszdurka: As discussed let's return promise and add then in http post handler.

lib/transactions.js
```diff
@@ -20,6 +20,20 @@ Transactions.prototype.find = function(id) {
 };
 
 /**
+ * @param {String} id
+ * @param {Object} message
+ */
+Transactions.prototype.execute = function(id, message) {
+  if ('ack' != message['janus']) {
```
tomaszdurka: `!==` :)

tomaszdurka:
Rest looks good.

vogdb:
@tomaszdurka please review.

lib/http-server.js
```diff
@@ -51,13 +52,30 @@ HttpServer.prototype.installRoutes = function(router) {
...
+    var streamId = String(params['streamId']);
+    var stream = streams.find(streamId);
+    if (stream) {
+      stream.proxyConnection.stopStream(stream.id)
+        .then(function() {
+          res.send({success: 'Stream stopped'});
+          return next();
+        }).catch(function(error) {
+          logger.warn('Stream stop failed', error);
+          res.send({error: 'Stream stop failed'});
```
tomaszdurka: Shall we send this message along?

vogdb: too technical

lib/http-server.js
```diff
@@ -51,13 +52,30 @@ HttpServer.prototype.installRoutes = function(router) {
 
   router.post('/stopStream', function subscribe(req, res, next) {
     var params = _.extend({}, req.params, req.body);
```
tomaszdurka: Let's not combine `params`, only `req.body`.

tomaszdurka:
Looks goo.

PR #19, opened 2015-11-05
Cleaning

fvovan:
- Optimise the excessive usage of `message.janus`. Store it in a variable where is needed.
- Replace all accesses to janus messages from `message.transaction` to `message['transaction']`.

@fvovan This one is a good starting point to keep up yourself with the changes.

### DIFF
--- a/lib/janus-proxy.js
+++ b/lib/janus-proxy.js
@@ -6,7 +6,6 @@ var ProxyConnection = require('./proxy-connection');
 
 var config = require('./config');
 var logger = require('./logger');
-var auth = require('./auth');
 
 /**
  * @param {Number} listenPort
--- a/lib/plugin/streaming.js
+++ b/lib/plugin/streaming.js
@@ -19,19 +19,21 @@ util.inherits(PluginStreaming, PluginAbstract);
  * @returns {Promise}
  */
 PluginStreaming.prototype.processMessage = function(message) {
-  if ('message' === message.janus && 'create' === message.body.request) {
+  var janusMessage = message['janus'];
+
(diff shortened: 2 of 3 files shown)


fvovan:
@vogdb please review

vogdb:
looks good. Some additions.
- Remove redundant usages of our `auth` module.
- Consider the changing of all message/response/request fields as the `transaction` field. It needs to be done to emphasize that they are data fields and not properties. Call me if you need an explanation on this.

fvovan:
@vogdb , and should `http-server.js` dots appearances  like
```logger.debug('request ' + req.path);
    if (req.body['apiKey'] != self.apiKey) {``` be converted too?
What is the rule of thumb to use dot notation? Actually i tried to not use it at all.

**Comment by vogdb** [comment]
> and should `http-server.js` dots appearances  like
```logger.debug('request ' + req.path);
    if (req.body['apiKey'] != self.apiKey) {``` be converted too?

`req.body` should not be changed to `req['body]` cause `body` is the property of `req`. If we had a class for janus messages like the below then we could have used dots. But we don't have it, so we ought to emphasize the fact that the accessed field can be absent on this object. 
```js
  function JanusMessage(){
    this.janus = ...,
    this.transaction = ...,
    //...
  }
```

vogdb:
@fvovan so basically you need to change dots to brackets everywhere we deal with janus messages. I see that there are still lots of dots usages. Should we pair on it?

fvovan:
No worries :) I will figure it out.

vogdb:
lgtm

PR #18, opened 2015-11-05
Feedback of unexpected/error cases.

vogdb:
For example.
-------------------------------------------------
```json
{
  "janus" : "error",
    "transaction" : "--------"
    "error" : {
      "code" : 405
      "reason" : "Illegal plugin to access"
    }
}
```
- ProxyConnection.onAttach:

  ```js
  if (!this.isAllowedPlugin(message.plugin)) {
    return Promise.reject('Disallowed plugin ' + message.plugin);
  }
  ```
- ProxyConnection.pluginFactory:

  ```js
    if (!pluginClass) {
      throw new Error('Invalid plugin');
    }
  ```

----------------------------------------------------------------------------------
```json
{
  "janus" : "error",
    "transaction" : "--------"
    "error" : {
      "code" : 403
      "reason" : "Unauthorized request"
    }
}
```

- BrowserConnection._onMessage:

  ```js
  auth.authorizeConnection(this, token)
  ```
- BrowserConnection._send:

  ```js
  auth.isValidConnection(this)
  ```

- PluginStreaming.onCreate:

  ```js
  auth.canPublish
  ```

- PluginStreaming.onWatch:

  ```js
  auth.canSubscribe
  ```

--------------------------------------------------------------

- ProxyConnection.onCreate. Just remove the error case.
- ProxyConnection.onDestroy. Remove it. Clear sessionId in hangup.

### DIFF
--- a/lib/proxy-connection.js
+++ b/lib/proxy-connection.js
@@ -63,17 +63,11 @@ ProxyConnection.prototype._processMessage = function(message) {
  * @returns {Promise}
  */
 ProxyConnection.prototype.onCreate = function(message) {
-  var self = this;
-  this.transactions.add(message.transaction, function(response) {
-    switch (response.janus) {
-      case 'success':
-        self.sessionId = response.data.id;
-        break;
-      case 'error':
-        //TODO Inform user that create didn't work. Should we close here the connection?
-        break;
+  this.transactions.add(message['transaction'], function(response) {
+    if ('success' == response['janus']) {
+      this.sessionId = response.data.id;
     }
-  });
+  }.bind(this));
   return Promise.resolve(message);
 };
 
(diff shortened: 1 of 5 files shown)


vogdb:
@tomaszdurka please review.
Some thoughts to discuss.
- Should we also pass an argument `message` to `JanusError`?
- Should we unify the error messages for `logger.error` and `JanusError`?
- I didn't transform `new Error('Invalid plugin')` at `ProxyConnection.prototype.pluginFactory` cause it does not relate to janus IMO. Wdyt about it?

@fvovan FYI

vogdb:
@tomaszdurka please review

lib/proxy-connection.js
```diff
@@ -19,7 +20,20 @@ function ProxyConnection(browserConnection, janusConnection) {
   this.janusConnection = janusConnection;
 }
 
+/**
+ * @param {String} message
+ */
 ProxyConnection.prototype.processMessage = function(message) {
+  return this._processMessage(message).catch(function(error) {
```
tomaszdurka: Can we use then catch on new lines? Also for another one? Like this:
```js
this._processMessage(message)
  .catch(function(error) {
  })
  .then(function(error) {
  });
```

tomaszdurka:
Rest looks good.

One thing to discuss - errors classes.
I feel strange that lower case `errors` stores multiple classes.
Maybe it should look rather like:
```js
var JanusError = require('janus-error');
throw new JanusError. Unauthorized(transaction);
```
@vogdb wdyt?

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka @fvovan  how do you think should we close the proxy connection if an errors happens?

vogdb:
paired with @tomaszdurka

PR #21, opened 2015-11-06
Rename apiKey to serverKey

fvovan:
Unified with CM part


fvovan:
@tomaszdurka please review. 
I found that config validation is not enough because empty value can be passed to HttpServer constructor by accident. Added a checking for it

tomaszdurka:
Didn't you miss config usage here https://github.com/fvovan/cm-janus/blob/issue-21/lib/cm-api-client.js?

fvovan:
true, my mistake. It's fixed

tomaszdurka:
Looks good.

PR #23, opened 2015-11-06
Add 'isPublish' attribute to Stream.

fvovan:
We need to know what actually stream represents (publish or subscribe). It is required for syncing, where we delete inconsistent data on CM part (non deleted for some reasons publisher streams, absent publishers streams for existing subscribers e t.c.)


fvovan:
@vogdb please review.

vogdb:
lgtm

PR #20, opened 2015-11-09
Create a standalone janus client for super-session connection.

vogdb:
Comes from cargomedia/janus-gateway-rtpbroadcast#14

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -49,6 +49,13 @@ Connection.prototype.isOpened = function() {
 };
 
 /**
+ * @returns {String}
+ */
+Connection.prototype.generateTransactionId = function() {
+  return Math.random().toString(36).substring(2, 12);
+};
+
+/**
  * @param {Object} message
  * @returns {Promise}
  * @private
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -49,6 +49,13 @@ Connection.prototype.isOpened = function() {
 };
 
 /**
+ * @returns {Number} webSocketStatus
+ */
+Connection.prototype.getStatus = function() {
+  return this.webSocket.readyState;
+};
+
+/**
  * @returns {String}
  */
 Connection.prototype.generateTransactionId = function() {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -11,6 +11,7 @@ var util = require('util');
  */
 function Connection(identifier, webSocket) {
   this.identifier = identifier;
+  this._waitingForResponse = false;
 
   this.webSocket = webSocket;
   this.webSocket.on('message', function(data) {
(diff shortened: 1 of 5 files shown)


vogdb:
@tomaszdurka please review and come up with a better name than `SuperUserSession` :)

vogdb:
@tomaszdurka please order the food

vogdb:
@tomaszdurka please don't forget about me

vogdb:
@tomaszdurka @tomaszdurka @tomaszdurka

vogdb:
:beetle_juice:

tomaszdurka:
:dancer:

lib/connection.js
```diff
@@ -81,6 +104,7 @@ Connection.prototype._send = function(message) {
  */
 Connection.prototype._onMessage = function(message) {
   try {
+    this._waitingForResponse = false;
```
tomaszdurka: This is a bit hacky, right? You might send 2 messages and after receiveing one you already set flag to false.
- This could cause some blocking state if you would send message which would not send anythiing back....?
- What about rather using transactions here? 
- Let's rather send keepalive only if there is no pending transaction?

tomaszdurka: More comment in the main comment :)

tomaszdurka:
Generally I like it. The only thing I don't like is (imo) overcomplicated `PeriodicTask` and `SuperSession` constructor.
Why not:
- Listen on `close` event and `connect` then.
- `connect` method should check if is connected, if yes return, otherwise do the connecting and run it with timeout.
- `keepalive` functionality could be done with `debounce`d method which would be run everytime super-session `send` and `receive` message.

vogdb:
> This is a bit hacky, right? You might send 2 messages and after receiveing one you already set flag to false.

It is not hacky. It just as simple as a hammer. If we don't receive anything after the last send then we lost the server. That's all =)

> - This could cause some blocking state if you would send message which would not send anythiing back....?

It will cause the reconnect. After we send any message, we must receive something back. That is how janus works.

> What about rather using transactions here? 

I thought about it. It will complicate code here cause `keepalive` messages are completed by `ack` response which we ignore currently. So we have to rework our transactions to either differentiate between usual transactions and keepalive transactions or we need to manually clear transactions and handle `ack` in all the transactions that we have.

> - Let's rather send keepalive only if there is no pending transaction?

Makes sense.


I propose to pair on this issue and finish it.

tomaszdurka:
Use `setInterval` and `_.debounce` for `PeriodicTask`s.

tomaszdurka:
As discussed further we will use different appraoch than websocket communication.
Closing in favor of #27

PR #24, opened 2015-11-10
AudioBridge plugin

vogdb:


lib/plugin/audio.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+PluginAbstract.prototype.onJoin = function(message) {
+  var self = this;
+  //message
+  self.proxyConnection.transactions.add(message['transaction'], function(response) {
+    self.stream = Stream.generate(message['body']['room'], self.proxyConnection);
+    logger.info('adding stream', self.stream);
+    streams.add(self.stream);
+  });
+  return Promise.resolve(message);
```
tomaszdurka: Was this to be comitted here? :panda_face:

lib/plugin/audio.js
```diff
@@ -0,0 +1,60 @@
...
+
+  return PluginAudio.super_.prototype.processMessage(message);
+};
+
+
+/**
+ * @param {Object} message
+ * @returns {Promise}
+ */
+PluginAbstract.prototype.onJoin = function(message) {
```
tomaszdurka: Oh, this is `PluginAbstract`, on purpose?

tomaszdurka:
Hmm, what about having a small trait-like/parent plugin called PluginStreaming. And use it in Audio/Video plugins?

vogdb:
@tomaszdurka what about having it as it is now

lib/plugin/audio.js
```diff
@@ -0,0 +1,60 @@
...
+      self.proxyConnection.transactions.add(message['transaction'], function(response) {
+        self.stream = Stream.generate(message['body']['id'], self.proxyConnection);
+        logger.info('adding stream', self.stream);
+        streams.add(self.stream);
+      });
+      return Promise.resolve(message);
+    }).catch(function() {
+      throw new JanusError.Unauthorized(message['transaction']);
+    });
+*/
```
tomaszdurka: Please delete commented code.

vogdb:
@tomaszdurka I think it makes sense to merge this only after #52 is merged

tomaszdurka:
@vogdb please merge upstream and adjust.

vogdb:
@tomaszdurka or @fvovan please review

vogdb:
I have a question. Should we have for Audio plugin `onCreate` handler? We have it for Video plugin stream but not Audio. However Audio plugin has `create` functionality ,i.e. janus client can create Audio streams(rooms).
cc @kris-lab @tomaszdurka

tomaszdurka:
We should have create callback handled. Please take #77 into consideration there.

vogdb:
@tomaszdurka this issue is quite overloaded already. Can we do `onCreate` separately cause it often painful to merge into this issue cause it changes Video plugin a lot.
So @tomaszdurka please review

tomaszdurka:
Looks good.

I will also move some method of plugin into session and will double-review then.

PR #22, opened 2015-11-10
Add readme

tomaszdurka:


README.md
```diff
@@ -1,2 +1,37 @@
-# cm-janus
-cm/janus bridge
+cm-janus
+========
+
+## About
+Bridge between cm-application and janus-gateway). Once running it should:
```
vogdb: Redundant bracket

README.md
```diff
@@ -1,2 +1,37 @@
...
+```
+cm-janus is based on single configuration file written in yaml format. Default config is present under [`bin/config.yaml`](bin/config.yaml).
+You can provide different config file using `-c` option (e.g. `bin/cm-janus -c /path/to/my/config/yaml`). New config will completely overwrite old one. Old one won't be used for defaults.
+
+Config format:
+
+```yaml
+logPath: 'log/app.log' # path to log file (relative to working dir)
+proxy:
+  listenPort: '8188' # port for incoming connections
```
vogdb: incoming websocket connections?

vogdb:
lgtm

tomaszdurka:
This is actually not 100% valid config with what we have currently. It's actually a roadmap how we should change some configuration options.

e.g.
- Use 2 diffrerent keys for cm-api incoming and outgoing requests.
- Move `logPath` level up (remove `app` namespace).

PR #25, opened 2015-11-10
Adjust config and split api keys.

fvovan:
Move proxy details to config and create split API keys for http-server  and cm-api-client

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Config format:
 ```yaml
 logPath: 'log/app.log' # path to log file (relative to working dir)
 proxy:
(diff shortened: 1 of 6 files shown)


fvovan:
@tomaszdurka  please review
so I renamed it to apiKey back but it will leave as 'serverKey' at CM's section.

tomaszdurka:
Looks ok I guess... we are so inconsistent in terms of how we pass params to classes.

fvovan:
hmm. do you mean that sometimes we cast config values but sometimes not and also we validate only few constructor arguments?

tomaszdurka:
Yes, it's about constructors, but let's make it in another issue.

PR #31, opened 2015-11-11
Consider using service manager concept for storing shared instances

tomaszdurka:


fvovan:
Generally it is okay, and I like the idea to pass config to App and then to use is values down there but I dislike what happened with logger. Now it is neither class nor singleton module. So it seems like tradeoff to avoid complexity of "full-featured DI".

Hmm, not sure about `ServiceManager.prototype.set = function(name, instance) {`
if we call passed function right in the beginning, what's the point to pass it?
So all our singleton instances will be initialised right after starting the app.

Then it's possible to instantiate Logger class (created beforehand :) ) at first lines of `Application.prototype.registerServices` and pass object to `set`

@tomaszdurka wdyt?

tomaszdurka:
Regarding the closure, it was simply to warp functionality connected with setting specific service. I think it could be stored when `set` and lazy/once executed within `get`.

There are couple of things I dislike in this solution. E.g. you cannot require any standalone `classes` without setting all needed for it services. Using `services.get` in head of the class file and not when it's used makes it like that. 
Alternative is to use `get` when you need service, not before (in start of the file), but that would complciate mocking and checking what is needed in order to instantiate a class (would require to scan file for usages, or run it to see exceptions). On the other hand sometimes you need/want only some dependencies.

------------------------------------
What about storing function passed into `set` method and execute it once when `get`ing? I like that more.

fvovan:
It definitely should be lazy. and moreover it will gives us possibility to resolve dependencies between services

tomaszdurka:
@fvovan please review

bin/cm-janus
```diff
@@ -1,17 +1,13 @@
...
-config.load(configPath);
-var configHash = config.asHash();
-
-var CMJanus = require('../lib/index.js');
+var config = Config.createFromFile(configPath).asHash();
 
-var proxy = new CMJanus.JanusProxy(String(configHash['proxy']['listenPort']), configHash['proxy']['janusAddress']);
-proxy.start();
+var application = new Application(config);
+application.registerServices();
```
fvovan: what if we put `registerServices()` inside its constructor?

tomaszdurka: Oh, I've just wanted this to be not auto-executable.
Maybe it should be rather:
```
var application = new Application();
application.registerServicesWithConfig(config);
```

lib/cm-api-client.js
```diff
@@ -1,6 +1,5 @@
-var logger = require('./logger');
+var serviceLocator = require('./service-locator')
```
fvovan: semicolon is ommitted

lib/config.js
```diff
@@ -92,4 +92,10 @@ Config._validate = function(config) {
   }
 };
 
-module.exports = new Config();
+Config.createFromFile = function(path) {
+  config = new Config();
```
fvovan: please add `var`

lib/janus-proxy.js
```diff
@@ -3,8 +3,7 @@ var JanusConnection = require('./janus-connection');
 var BrowserConnection = require('./browser-connection');
 var ProxyConnection = require('./proxy-connection');
 
-var config = require('./config');
-var logger = require('./logger');
+var serviceLocator = require('./service-locator')
```
fvovan: please add semicolon

tomaszdurka: Did scan on semicolons, should be ok now.

bin/cm-janus
fvovan: For me second way is worse, `config` can have some params not related to services.

fvovan:
rest looks good

tomaszdurka:
Ok, let's keep registerServices as is for now.

PR #27, opened 2015-11-12
Implement Job Manager

vogdb:
- based on filesystem
- uses specific folder for all the jobs information
- job information is in json format
```json
{
 "plugin": "rtpbroadcast",
 "event": "recording",
 "data": {
  "streamChannelId": "1",
  "audio": "/path/123.rtp",
  "video": "/path/124.rtp"
 }
}
```
Note: all the files provided in data are ready, job can be executed straight away

### Workflow
- should read all json-files and start processing them
- should start listening for all new json-files and process them
- process: identify and execute specific job with `data`
- should remove json-file when complete

### Implementation details:
- use https://github.com/c4milo/node-inotify for listening for file events


vogdb:
@tomaszdurka please review. Please pay attention to TODO items.

fvovan:
btw, why we can't use something proven for message queues like rabbitMQ or beanstalk?
I think it could be less problematic but better scalable

tomaszdurka:
@fvovan Well message queues were our first choice as they are more flexible. Filesystem has one main advantage that we don't need to care about disconnects/reconnects and this way it's a bit more reliable. We have luxury of using filesystem as we know it all will stay on single machine.

fvovan:
And will it be just single instance? Frankly not very familiar with nodejs deployment schemes and performance limitations but saw that a rule of thumb is running instances equals to cores amount.

vogdb:
@tomaszdurka what do you think?

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+var serviceLocator = require('../service-locator');
+var logger = serviceLocator.get('logger');
+var Job = require('./job');
+var TestJobHandler = require('./test-job-handler');
+
+/**
+ * @param {String} filepath, an Absolute one!
+ * @constructor
+ */
+function JobManager(filepath) {
```
tomaszdurka: `filepath` rather call it `jobsDirectory(Path)`?

tomaszdurka: Not sure.

tomaszdurka: What I meant is, it's not really a file, right?

vogdb: ok

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+JobManager.prototype._installFileListeners = function() {
+  this._watch = this._inotify.addWatch({
+    path: this._filepath,
+    watch_for: Inotify.IN_CLOSE_WRITE,
+    //watch_for: Inotify.IN_AvaLL_EVENTS,
+    callback: function(event) {
+      var mask = event.mask;
+      var isFile = !(mask & Inotify.IN_ISDIR);
```
tomaszdurka: Maybe we should do a few more checks: is file, is readible, is json, has proper keys; or not them at all?

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+  });
+};
+
+JobManager.prototype._handleJobFile = function(filepath) {
+  fs.readFile(filepath, function(error, data) {
+    if (error) {
+      logger.error('Could not read the file "' + filepath + '"', error);
+      //TODO should we retry the attempt here? It would be silly to throw or stop the process because of one failed file
+    }
+    try {
```
tomaszdurka: Oh, let's do the parsing outside of the job.

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+      var job = new Job(data);
+      var handler = this._jobHandlers[job.type];
+      handler.handle(job);
+    } catch (e) {
+      logger.error('Could not handle the file "' + filepath + '"', e);
```
tomaszdurka: This is very specific message.
What do you think about dividing all the messages into two kinds (coming from two checks): 
- invalid file (no file, can't read, invalid json)
- can't handle job (some internal problems with the data, jobhandler configured badly?)

lib/job/test-job-handler.js
```diff
@@ -0,0 +1,12 @@
+function TestJobHandler() {
```
tomaszdurka: Hmm what about rather having specific job classes, and single job handler which would find proper job and execute it?

tomaszdurka: Actually there no much difference. But I would drop the whole Job class.
Basically have something to match json into specific job handler (or job). Then only pass data into that handler.

lib/job/job-manager.js
vogdb: Not at this level. Here I do the check for Directory only cause otherwise `event.name` property is `undefined`. See the docs of Inotify https://github.com/c4milo/node-inotify#event-object-structure
> The event.name property is only present when an event is returned for a file inside a watched directory; it identifies the file path name relative to the watched directory.

lib/job/test-job-handler.js
vogdb: ok

lib/job/job-manager.js
vogdb: ok

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+JobManager.prototype._handleJobFile = function(filepath) {
+  fs.readFile(filepath, function(error, data) {
+    if (error) {
+      logger.error('Could not read the file "' + filepath + '"', error);
+      //TODO should we retry the attempt here? It would be silly to throw or stop the process because of one failed file
+    }
+    try {
+      var job = new Job(data);
```
tomaszdurka:
```
var handler = this.findHandler(data['plugin'], data['eventName'])
handler.handle(data['data']); // {videoFile: '/foo', audioFile: '/bar'}
```

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+JobManager.prototype.reset = function() {
+  this._removeFileListeners();
+  this._installFileListeners();
+};
+
+JobManager.prototype._setupJobHandlers = function() {
+  this._jobHandlers = {};
+  [TestJobHandler].forEach(function(handlerClass) {
+    var handler = new handlerClass();
+    this._jobHandlers[handler.getType()] = handler;
```
tomaszdurka: `plugin + ':' + eventName`

lib/job/job-manager.js
```diff
@@ -0,0 +1,76 @@
...
+var Job = require('./job');
+var TestJobHandler = require('./test-job-handler');
+
+/**
+ * @param {String} filepath, an Absolute one!
+ * @constructor
+ */
+function JobManager(filepath) {
+  var fileStats = fs.statSync(filepath);
+  if (!fileStats.isDirectory()) {
```
tomaszdurka: You can inline this :)

vogdb:
@tomaszdurka please review.

lib/job/job-manager.js
```diff
@@ -0,0 +1,96 @@
...
+    .then(function(job) {
+      var handler = this._getJobHandler(job['plugin'], job['event']);
+      handler.handle(job['data']);
+    }.bind(this))
+    .catch(function(error) {
+      logger.error(error);
+    });
+};
+
+JobManager.prototype._readJob = function(filepath) {
```
tomaszdurka: Let's be more explicit `_readJobFile`?

lib/job/job-manager.js
```diff
@@ -0,0 +1,96 @@
...
+    fs.readFile(filepath, function(error, content) {
+      if (error) {
+        reject(new Error('Invalid file "' + filepath + "'. " + error.message));
+      }
+      try {
+        var job = JSON.parse(content);
+      } catch (e) {
+        reject(new Error('Invalid file content "' + filepath + "'. " + e.message));
+      }
+      if (!job['plugin'] || !job['event'] || !job['data']) {
```
tomaszdurka: We don't filter out other keys which might be put there and someone thought they will be parsed/used.


What about rather:
```js
var currentKeys = _.keys(job);
var validKeys = ['plugin', 'event', 'data'];
if (_.isEqual(validKeys, currentKeys)) {
  reject(new Error('Invalid or missing keys: ' + currentKeys.join(', ') + ', allowed/required keys are: ' + validKeys,join(',')));
}
```
?

Btw I've played with underscore and came up with something like this....
```
_.every([1,2], _.contains.bind(_, [1,2,3]));
```
...

lib/job/job-manager.js
```diff
@@ -0,0 +1,96 @@
...
+var logger = serviceLocator.get('logger');
+var TestJobHandler = require('./test-job-handler');
+
+/**
+ * @param {String} jobDirectory, an Absolute one!
+ * @constructor
+ */
+function JobManager(jobDirectory) {
+  if (!fs.statSync(jobDirectory).isDirectory()) {
+    throw new Error('`filepath` must be a directory');
```
tomaszdurka: Let's be consistent. `jobDirectory`? or `path?

lib/job/job-manager.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+  this._filepath = jobDirectory;
```
tomaszdurka: same as about `directory`/`path`?

lib/job/test-job-handler.js
tomaszdurka: Maybe still put all these into `job/handler` directory. Plus add `abstract.js` to know which methods need to be implemented?

tomaszdurka:
Review done.

lib/job/job-manager.js
vogdb: Lets not be so strict here. There is no need for that. The required fields are enough.

vogdb:
@tomaszdurka please review

lib/job/job-manager.js
tomaszdurka: What if there will be extra field?

tomaszdurka: Silently ignore?

vogdb: it is better than reject the handling of a correct file

tomaszdurka:
Couple of things still:
- Registry uses handlers list which cannot be modified at all. What about making it public and run `registerHandlers()` from outside (`JobManager`).
- We are still missing parsing existing files. 
- We are still missing deleting the file after job has been done (on error too?)

vogdb:
> - Registry uses handlers list which cannot be modified at all. What about making it public and run `registerHandlers()` from outside (`JobManager`).

That is fine. It just encapsulates the registering in itself.

Agree with the rest.

vogdb:
> - We are still missing deleting the file after job has been done (on error too?)

Should not we keep error files for investigation?

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
tests are failing cause of outdated node version

tomaszdurka:
As agreed:
- let's remove TestJobHandler
- use AbstractJobHandler mock

vogdb:
@tomaszdurka please review

vogdb:
also what about tests? Should we wait here for the fix of node.js version ?

tomaszdurka:
Let's add inotify here, at least we check if tests are passing.

tomaszdurka:
Waiting on nodejs upgrade for Vagrant box. Otherwise dev environment won't work.

vogdb:
@tomaszdurka merge?

tomaszdurka:
@vogdb merging after small adjustments

PR #38, opened 2015-11-12
make Logger to be a class

fvovan:
Possibly try to wrap existing code with new created class and delegate all used methods to minilog.

### DIFF
--- a/bin/config.yaml
+++ b/bin/config.yaml
@@ -1,5 +1,4 @@
-app:
-  logPath: 'log/app.log'
+logPath: 'log/app.log'
 httpServer:
   port: 8888
   apiKey: '123fish'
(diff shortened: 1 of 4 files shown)


fvovan:
@vogdb please review.

vogdb:
Can we somehow get rid of duplicating of `logger` methods?
Just thinking.
```js
var logger = new Logger(__dirname + '/../' + this.config['logPath'], 'app');
return logger.getLogpipe();
```
```js
Logger.prototype.getLogpipe = function() {
  return this._logger;
};
```
not sure through, you can have your ways

fvovan:
@vogdb please review.
yep I thought about it, downside is we now are binded to minilog as we directly use its methods.

vogdb:
lgtm

PR #42, opened 2015-11-16
Make Logger more oop

tomaszdurka:
Allow to create instances of and mock Logger class.

Service registry:
```js
var Logger = require('./logger');
var serviceLocator = require('./service-locator');
serviceLocator.register('logger', function() {
  var logger = new Logger();
  var logFilePath = __dirname + '/../' + this.config['logPath']
  var stream = fs.createWriteStream(logFilePath, {flags: 'a', defaultEncoding: 'utf8'});
  logger.pipe(stream);
  return logger;
});
```

Not sure if it's possible with minilog which signature is quite strange (unless you know different way than we use).

### DIFF
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -32,12 +32,12 @@ CMApiClient.prototype._request = function(action, data) {
   serviceLocator.get('logger').info('cm-api', 'request', options.uri, options.body);
   return requestPromise(options)
     .catch(function(error) {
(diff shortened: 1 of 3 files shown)


tomaszdurka:
Paired with @fvovan.\
@fvovan please review!

fvovan:
lgtm

vogdb:
I see you like to work in pair ^____^

tomaszdurka:
^^

PR #43, opened 2015-11-17
Connection unit tests

vogdb:
Based on #27


test/helpers/websocket.js
```diff
@@ -0,0 +1,39 @@
+require('mock-socket');
```
tomaszdurka: Isn't this missing some declaration?

vogdb: Unfortunately no. Specifics of this library :( It imports to global.

vogdb:
@tomaszdurka please review

tomaszdurka:
So this bascically tests that messages are forwaded within connection. Looks good for me.
We still don't test things like queuing.
Let's merge this one though.

PR #45, opened 2015-11-17
CmApiClient unit tests

vogdb:
depends on #43


test/unit/cm-api-client.js
```diff
@@ -0,0 +1,71 @@
+var expect = require('chai').expect;
```
tomaszdurka: I'm rather fan of assert then expect, could we use one or there are reasons to use each of them for different cases? :panda_face:

test/unit/cm-api-client.js
```diff
@@ -0,0 +1,71 @@
...
+var nock = require('nock');
+
+var App = require('../../lib/index');
+new App({}).registerServices();
```
tomaszdurka: This will now cause problems (exceptions), as we use `path.resovle`.

Anyway I think this behaviour created `undefined` file in project root dir. I've seen it couple of times and was not sure from where. It basically logged into file `__dirname + '../' + undefined`...

test/unit/cm-api-client.js
vogdb: BDD test style is considered to be more future-proof. That is why I picked it (expect/should).

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
@fvovan please review

fvovan:
lgtm

PR #46, opened 2015-11-17
Make cm-janus runnable in VM

ppp0:

PR #47, opened 2015-11-17
Use node v0.12 on debian / vm

ppp0:


ppp0:
Closes #46

puppet/Puppetfile
```diff
@@ -5,3 +5,7 @@ mod 'cargomedia/nodejs',
 mod 'cargomedia/build',
   :git => 'git@github.com:cargomedia/puppet-packages.git',
   :path => 'modules/build'
+
+mod 'cargomedia/apt',
+  :git => 'git@github.com:cargomedia/puppet-packages.git',
+  :path => 'modules/apt'
```
tomaszdurka: Is this needed? Shouldn't nodejs already include `apt` module in it's puppetfile?

ppp0:
Agree

Somehow it was not happening?

But I cross-checked and it's being installed now

PR #48, opened 2015-11-17
Allow to use arbitrary paths for log

tomaszdurka:
After having tried to run as a non-root user I found out that the log location needs be in a place where the app user has rights to write to

Ideally, this is configurable. A common place where log files are stored in our setup is `/var/log/<appname>`

As-of-now, this is only possible by prefixing the path with as many `../` as the depth of where the app is installed, e.g. `logPath: '../../../../../../../../home/vagrant/app.log'` because of https://github.com/cargomedia/cm-janus/blob/master/lib/index.js#L19

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,6 +4,7 @@ var Auth = require('./auth');
 var Streams = require('./streams');
 var Logger = require('./logger');
 var fs = require('fs');
+var path = require('path');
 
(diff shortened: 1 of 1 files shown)


tomaszdurka:
@vogdb please review and merge

vogdb:
lgtm

PR #44, opened 2015-11-18
Add CMApiClient tests

fvovan:
- test that requests are created properly
- test that client interprets properly responses - certain types (successes, errors on app and http level)


fvovan:
@tomaszdurka @vogdb  please take a look
I did it mainly via mocking `request-promise`

vogdb:
Actually the tests themselves are good.
- remove describe nesting and never use it.
- Fix `serviceLocator.register`. Cause now there is a problem. For example lets take `_request` testsuite. Currently it fails, but if we change our approach with `loggerSpy` and `serviceLocator.register`. It works fine.
```js
  describe('_request()', function() {
    this.timeout(2000);
    var baseUri = 'http://cm.dev/';
    var apiKey = 'fooKey';
    var cmHttpClient = new CMApiClient(baseUri, apiKey);
    var serviceLocator = require('../../lib/service-locator.js');

    var logger = {
      info: sinon.stub()
    };

    var loggerSpy = logger['info'];
    serviceLocator.register('logger', logger);

    var action = 'foo';
    var sentData = ['bar', 'baz'];
```

 - There is one big IMO. We should not test the calls to `log`. That is too much. So, no `assert.isTrue(loggerSpy.calledTwice);`. Very very too much.

vogdb:
It can be fixed even simpler, if we call to the just registered service immediately.
```js
    var loggerSpy = sinon.spy(logger, 'info');
    serviceLocator.register('logger', function() {
      return logger;
    });
    serviceLocator.get('logger');
```

vogdb:
hm, still seems to be the problem of singletons in the end :(

vogdb:
Also there is a problem in `Connection` tests. We should not register services right after require. We must use the appropriate test structure, e.g. `before` and etc.

vogdb:
My suggestion is that every test should create its mocks in `before` and clean them in `after`! The last part is especially important.

fvovan:
@vogdb  please review

vogdb:
looks fine

PR #49, opened 2015-11-18
Cleanup tests

fvovan:
- Use `assert` in all tests to be consistent. Including the one that are not merged. FYI @tomaszdurka.
- Rename jobManager => job-manager. Delete all obsolete `require` in it.


fvovan:
@vogdb please review,
https://github.com/cargomedia/cm-janus/pull/45 and 
https://github.com/cargomedia/cm-janus/pull/43 cleanings are also done.

vogdb:
looks good. One thing in #45 `serviceLocator.register('logger', new Logger());` I guess should be changed too.

test/spec/job-manager.js
```diff
@@ -8,7 +8,9 @@ var JobManager = require('../../lib/job/job-manager');
 var AbstractJobHandler = require('../../lib/job/handler/abstract');
 var Logger = require('../../lib/logger');
 var serviceLocator = require('../../lib/service-locator');
-serviceLocator.register('logger', new Logger());
+serviceLocator.register('logger', function() {
+  return new Logger();
+});
```
tomaszdurka: You don't need to make it lazy.

tomaszdurka: Matter of preference I guess.

fvovan: Not completely,
Mocha runs tests in a parallel, so you have the same ServiceLocator per each test.
And if you register service by value (not by  function) then it's impossible to override it with function, as it checks if value is defined first and then calls function. Can be changed later but just non dirty fix for now

PR #57, opened 2015-11-19
Add simple unit test for service-locator

fvovan:


fvovan:
@vogdb please review.
I tried to do impletend calls counter via sinon's spies, but don't know why it didn't work  for me (spy was not attached).

vogdb:
- lets be consistent with names. Somewhere you cut names, somewhere don't. Lets not cut them at all. For example, `funKey` and `funVal`.
- Why there is a second arg in calls `serviceLocator.get(funKey, serviceFunction)` ? ServiceLocator.get doesn't have only one arg? `ServiceLocator.prototype.get = function(name) {`?

fvovan:
@vogdb please review

vogdb:
- lets use `assert(serviceSpy.calledOnce);` instead of callCount
- You are right. `after` is redundant. Let tests to clean themselves manually. In that case we can also move variables from closure scope into the tests itself.

fvovan:
I moved cleaning to end of `it` block

vogdb:
looks very good to me!

PR #52, opened 2015-11-19
Add tests for proxy-connection

vogdb:


vogdb:
@fvovan please review. Couldn't think of unit tests here.

vogdb:
@fvovan please postpone

vogdb:
@fvovan please review

fvovan:
looks good, but I would add cases for failures. For example for not allowed plugin or other cases when `_processMessage` is rejected.

vogdb:
@fvovan please review.
@tomaszdurka please look as well. As you can see I've changed completely how transactions work. Now `transaction.execute` returns Promise. IMO it was a bug that before transactions were not able to control workflow of `processMessage`. Now transaction is able resolve/reject the processing message. Also now `processMessage` waits for finishing of transaction cause it can take some time (see `PluginStreaming.onCreate` for example of long async transaction).

Also there is only one test for error response (attach of illegal plugin). There are no tests for error during create of session or attach of plugin cause currently we don't handle them and just passing them further to janus client. I didn't write any tests for plugin direct methods like: `getPlugin`, `removePlugin`, - cause I believe that we need to remove this part from `ProxyConnection`. The issue is already created for this here #58.

lib/transactions.js
```diff
@@ -22,14 +22,24 @@ Transactions.prototype.find = function(id) {
...
+      return Promise.resolve();
+    case 'error':
       this.remove(id);
-    }
+      return Promise.resolve();
+    default:
+      var transaction = this.find(id);
+      if (transaction) {
+        return transaction.call(null, message).then(function() {
+          this.remove(id);
```
tomaszdurka: This should be in finally?

lib/proxy-connection.js
```diff
@@ -29,13 +29,14 @@ ProxyConnection.prototype.processMessage = function(message) {
       error = new JanusError.Unknown(message['transaction']);
     }
     this.browserConnection.send(error.response);
+    throw error;
```
fvovan: Do we need to rethrow it here?

test/spec/proxy-connection.js
```diff
@@ -0,0 +1,169 @@
...
+        sinon.stub(proxy, 'close');
+        var destroyRequest = {
+          janus: 'destroy',
+          transaction: proxy._generateTransactionId()
+        };
+        var destroyResponse = {
+          transaction: destroyRequest.transaction
+        };
+
+        assert(!proxy.close.calledOnce);
```
fvovan: `!proxy.close.called` would be better here

test/spec/proxy-connection.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    proxy.processMessage(attachRequest).then(function() {
+      proxy.processMessage(attachResponse).then(function() {
+        assert(proxy.getPlugin(attachResponse.data.id) instanceof PluginStreaming);
+      });
+    });
+  });
+
+  it.only('Attach illegal plugin', function() {
```
fvovan: is `only()` needed here? afaik Mocha executes only test marked with this block, right?

lib/proxy-connection.js
vogdb: Yes, otherwise we resolve here and pass the message further. There is no need to pass the message further cause we already informed `browserConnection` about this error and it should behave accordingly on its own, I mean the `browserConnection`.

vogdb:
@tomaszdurka @fvovan please review

tomaszdurka:
Consider plugin erorrs `(message.janus == 'event' && message.body.error)`

vogdb:
@tomaszdurka please review

tomaszdurka:
Let's merge this!

PR #59, opened 2015-11-19
Inject ServiceLocator to application classes

fvovan:


fvovan:
@vogdb , @tomaszdurka could you have a look ?
Frankly speaking for me it's quite a bit messy and weird. Also it requires better control for constructor params (checking type).
Also I have another idea how to deal with it. But I'm not sure how to do it in node now. Injection could be optional if we somehow export not only `new ServiceLocator()` but also classname itself. Then we could move SL to last constructor param make it optional and check in constructor if it was provided. If yes - set it to attribute, if not - use global singleton. Or would it be better just use tests walker?

vogdb:
No no no no no :)

vogdb:
Ok. We are going to use your last approach as this:
```js
var instance = new ServiceLocator();
ServiceLocator.getInstance = function() {
  
};

module.exports = ServiceLocator;
```

Also I propose to drop `service-locator` and just implement this Singleton pattern for every service need. For example, Auth:
```js
var instance = new Auth();

Auth.getInstance = function(){
  return instance;
};

module.exports = Auth;
```

Also, we can upgrade this pattern in the next way:
```js
var _instance;

Auth.getInstance = function(){
  return _instance;
};
Auth.setInstance = function(instance){
  _instance = instance;
};

module.exports = Auth;
```

So we just need to set it once in `index` file and after use it. And for tests we can replace it easily.

@tomaszdurka @fvovan

fvovan:
afaiu in case of `var` all instances of needed class will again have the same `_instance` value (e.g. singleton). And that's the main problem. How we replace it for particular test, not touching  other running in the same tests if it is actually the same instance?

vogdb:
Cause we export the classname, we can easily create a new instance in tests.
```js
var Auth = require('./Auth'); // classname
var singleton = Auth.getInstance(); // singleton instance
var authInstance = new Auth(); // completely new instance, independent of singleton
Auth.setInstance(authInstance); // we can even replace it
```

fvovan:
I understand that we can create instances but what is the way to use only the instance we need?
 but let's consider 2 tests create 2 instances of Auth (or another class) and store them via `setInstance()` method.
The only way to get instance in user class is 
```var Auth = require('./Auth').getInstance;```
What actually instance it will be?

**Comment by vogdb** [comment]
Cause mocha tests are executed [sequentially](https://github.com/mochajs/mocha/issues/321#issuecomment-4541749). And there are no plans to make it async mochajs/mocha#98. I think we can do smth like this:
```js
before(function(){
  var a = new Auth();
  Auth.setInstance(a);
});
after(function(){
  Auth.setInstance(null);
});
```
so, the `a` instance will be used in all tests of one test file.

tomaszdurka:
Guys, let's not worry about Auth please. I think we might remove it all together some time soon.
Regarding service-locator I would keep it as is! @fvovan All of this is done to make it testable or basically more "correct"? We should be able to leave with single global singelton for now. 

Plus testing it ain't so hard. It doesn't have any state we cannot really check (it's javascript) and doesn't have any state set in constructor too.

vogdb:
There is no need in refactoring of the current `serviceLocator`. It works fine. Please see https://github.com/cargomedia/cm-janus/pull/44#issuecomment-158380613

tomaszdurka:
Closable?

fvovan:
yesss

PR #62, opened 2015-11-20
Add cm-binary path to config

tomaszdurka:


tomaszdurka:
@vogdb please review

This will be used by job-managers. I add this to include this already in our puppet manifests.

vogdb:
lgtm

vogdb:
hmm, should not we merge those `cm` properties?

vogdb:
like `cmApi`, `cmApplication`?

tomaszdurka:
Let's leave for now. Depends what kind of separation we want in config.
- CM application is about local application.
- CM api about remote machine.

Let's leave separate - we can merge anytime.

PR #53, opened 2015-11-20
Add tests for streams

tomaszdurka:


tomaszdurka:
@fvovan please review

test/unit/streams.js
```diff
@@ -0,0 +1,69 @@
...
+var ProxyConnection = require('../../lib/proxy-connection');
+var sinon = require('sinon');
+var _ = require('underscore');
+
+
+describe('streams', function() {
+
+  it('add', function() {
+    var streams = new Streams();
+    var stream = new Stream('foo', 'bar', sinon.mock(ProxyConnection));
```
fvovan: I think it's not a good idea to use mock
you can try 
>If you want to create a stub object of MyConstructor, but don’t want the constructor to be invoked, use this utility >function.
```
var stub = sinon.createStubInstance(MyConstructor)
```

test/unit/streams.js
```diff
@@ -0,0 +1,69 @@
...
+
+    assert.strictEqual(_.size(streams.list), 1);
+    assert.equal(streams.list['foo'], stream);
+  });
+
+  it('remove', function() {
+    var streams = new Streams();
+    var stream = new Stream('foo', 'bar', sinon.mock(ProxyConnection));
+    streams.list = {foo: stream};
+    assert.strictEqual(_.size(streams.list), 1);
```
fvovan: You can use 
>`assert.lengthOf([1,2,3], 3, 'array has length of 3');`

test/unit/streams.js
```diff
@@ -0,0 +1,99 @@
...
+    var stream = sinon.stub();
+    stream.id = 'foo';
+    streams.list[stream.id] = stream;
+    assert.strictEqual(streams.find('foo'), stream);
+    assert.strictEqual(streams.find('bar'), null);
+  });
+
+  it('findAllByConnection', function() {
+    var streams = new Streams();
+    var connection1 = sinon.mock(ProxyConnection);
```
fvovan: Not sure about `mock` usage here

tomaszdurka:
@fvovan please review

I've removed my own `throw` implementation.
Replaced `stub`/`mock` with `createStubInstance`

tomaszdurka:
@fvovan added checks about instance types

test/unit/streams.js
```diff
@@ -0,0 +1,96 @@
+var assert = require('chai').assert;
+var Stream = require('../../lib/stream');
+var Streams = require('../../lib/streams');
+var ProxyConnection = require('../../lib/proxy-connection');
+var sinon = require('sinon');
+var _ = require('underscore');
+
+assert.equalArray = function(array1, array2) {
+  assert.instanceOf(array1, Array);
```
fvovan: I think it's okay but `typeOf` could be used for standard types

test/unit/streams.js
```diff
@@ -0,0 +1,96 @@
...
+describe('streams', function() {
+
+  it('add', function() {
+    var stream1 = sinon.createStubInstance(Stream);
+    stream1.id = 'foo';
+    var stream2 = sinon.createStubInstance(Stream);
+    stream2.id = 'bar';
+
+    var streams = new Streams();
+    assert.strictEqual(_.size(streams.list), 0);
```
fvovan: `lengthOf` could be more clear

fvovan:
looks good with some minors

test/unit/streams.js
tomaszdurka: Let's keep this.

test/unit/streams.js
tomaszdurka: It's actually object, length doesn't work :(

fvovan: I meant
`assert.lengthOf(object, length, [message])`
didn't actually check it with objects but it says 
>Asserts that object has a length property with the expected value.

tomaszdurka: This throws when used on object. It checks length prop.

PR #63, opened 2015-11-20
Global error handler in tests

vogdb:
Currently if an error is thrown inside a promise without `catch` block, we loose it and tests exit by timeout.


vogdb:
@tomaszdurka please review
http://bluebirdjs.com/docs/api/error-management-configuration.html#global-rejection-events

PR #51, opened 2015-11-21
Add tests for http-server

fvovan:


fvovan:
@vogdb could you review please.
Did I use promises properly? Tried to insert sinon's spies calls assertions to `finally` block, but for some reason all spies and stubs got out of  'working scope'.

vogdb:
@fvovan  Please fix all TODO items and fix all usages of spies. Also please research the changes that are made. Currently you are using the same spies/stubs for all requests without resetting stubs' state. It means that on the second test the `findSpy` will be already `calledOnce` and the second test will make it `calledTwice`.

vogdb:
lgtm

PR #58, opened 2015-11-23
Move pluginFactory out of ProxyConnection

vogdb:
Right after #52 is completed

### DIFF
--- a/lib/plugin/plugin-registry.js
+++ b/lib/plugin/plugin-registry.js
@@ -1,0 +1,42 @@
+/**
+ * @param {Array} pluginClassList
+ * @constructor
+ */
+function PluginRegistry(pluginClassList) {
+  this._types = {};
+  pluginClassList = pluginClassList || [];
+  pluginClassList.forEach(function(plugin) {
+    this.registerPlugin(plugin);
+  }.bind(this));
+}
+
+/**
+ * @param {Function} pluginClass
+ */
+PluginRegistry.prototype.registerPlugin = function(pluginClass) {
+  if (!(pluginClass instanceof Function)) {
+    throw new Error('Plugin "' + pluginClass.constructor + '" must be instantiable');
+  }
+  if (!pluginClass.TYPE) {
+    throw new Error('Plugin "' + pluginClass.constructor + '" must have a TYPE static field');
+  }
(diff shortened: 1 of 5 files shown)


vogdb:
@fvovan please review

lib/plugin/plugin-registry.js
```diff
@@ -0,0 +1,42 @@
...
+  pluginClassList.forEach(function(plugin) {
+    this.registerPlugin(plugin);
+  }.bind(this));
+}
+
+/**
+ * @param {Function} pluginClass
+ */
+PluginRegistry.prototype.registerPlugin = function(pluginClass) {
+  if (!(pluginClass instanceof Function)) {
```
fvovan: Shouldn't it be an instance of `PluginAbstract` ?

lib/plugin/plugin-registry.js
```diff
@@ -0,0 +1,42 @@
...
+PluginRegistry.prototype.isAllowedPlugin = function(type) {
+  return !!this._types[type];
+};
+
+PluginRegistry.prototype.createPlugin = function(type, id) {
+  var pluginClass = this._types[type];
+  if (!pluginClass) {
+    throw new Error('Invalid plugin');
+  }
+  return new pluginClass(id, type, this);
```
fvovan: Third argument of  `PluginAbstract` is `proxyConnection` but here you pass `PluginRegistry ` as `this`
> `function PluginAbstract(id, type, proxyConnection)`

lib/plugin/plugin-registry.js
vogdb: It's long to explain. The code shows it better
```js
var util = require('util');

function A() {
}
function B() {
  A.apply(this, arguments)
}
util.inherits(B, A);

console.log(B instanceof A);// false
var b = new B();
console.log(b instanceof A);// true
```

In `registerPlugin` we are waiting for class not instance of class

fvovan: Thanks. I realised it. Actually I updated my comment but probably it was too late.

fvovan:
the rest looks okay

vogdb:
@fvovan please review

vogdb:
@fvovan please review

fvovan:
lgtm

PR #55, opened 2015-11-23
Refactor streaming plugin

vogdb:


vogdb:
@fvovan please review.
All error cases will be covered in #66. All error processing of plugins should be changed.
@tomaszdurka FYI

test/unit/plugin-video.js
```diff
@@ -0,0 +1,197 @@
...
+var ProxyConnection = require('../../lib/proxy-connection');
+var PluginVideo = require('../../lib/plugin/video');
+
+var Logger = require('../../lib/logger');
+var CmApiClient = require('../../lib/cm-api-client');
+var Stream = require('../../lib/stream');
+var Streams = require('../../lib/streams');
+var serviceLocator = require('../../lib/service-locator');
+
+describe.only('Video plugin', function() {
```
fvovan: Please remove `only`

test/spec/proxy-connection.js
```diff
@@ -5,7 +5,7 @@ var EventEmitter = require('events');
 require('../helpers/global-error-handler');
 var JanusError = require('../../lib/janus-error');
 var ProxyConnection = require('../../lib/proxy-connection');
-var PluginStreaming = require('../../lib/plugin/streaming');
+var PluginStreaming = require('../../lib/plugin/video');
```
fvovan: I would rename `PluginStreaming` too

test/unit/plugin-video.js
vogdb: Ahhhh, I'm so bad at this

test/unit/plugin-video.js
```diff
@@ -0,0 +1,197 @@
...
+
+  this.timeout(2000);
+
+  before(function() {
+    serviceLocator.reset();
+    serviceLocator.register('logger', new Logger());
+    serviceLocator.register('cm-api-client', function() {
+      var baseUrl = 'http://localhost:8080';
+      var apiKey = 'apiKey';
+      var mockRequest = function() {
```
fvovan: why we need this `mockRequest`  with dummy server? I don't actually understand what is does.
wouldn't it be better to do something like 
``` var apiClient = new CmApiClient(baseUrl, apiKey);
apiClient._request = function () {
 return 'Created'; // or what it expects and even can stub it
}
return apiClient;
```

vogdb: Well. I didn't want to stub private methods, so I just mocked the success responses from server.

fvovan: Does it require to use recursion ?

vogdb:
@fvovan please review

fvovan:
lgtm

PR #68, opened 2015-11-23
Add code-coverage

tomaszdurka:


tomaszdurka:
@vogdb @fvovan please review

vogdb:
lgtm

vogdb:
@tomaszdurka is that ok?
```
/home/vagrant/cm-janus/node_modules/codecov.io/lib/getConfiguration.js:27
      throw new Error("unknown service. could not get configuration");
            ^
Error: unknown service. could not get configuration
    at module.exports (/home/vagrant/cm-janus/node_modules/codecov.io/lib/getConfiguration.js:27:13)
    at sendToCodecov (/home/vagrant/cm-janus/node_modules/codecov.io/lib/sendToCodeCov.io.js:11:15)
    at Socket.<anonymous> (/home/vagrant/cm-janus/node_modules/codecov.io/bin/codecov.io.js:14:5)
    at Socket.emit (events.js:129:20)
    at _stream_readable.js:908:16
    at process._tickCallback (node.js:355:11)
```

fvovan:
The same problem for me as well

tomaszdurka:
This is when you run it locally. You could configure it to work locally by providing configuration. In our case it's automatically wired on github, and we only need it there.
I could also add something like `test-with-coverage` for running this locally.

vogdb:
lgtm then

PR #28, opened 2015-11-24
Implement rtpbroadcast recording job handler

fvovan:
Receives data:
```json
{
  "streamChannelId": "id",
  "audio": "/path/123.rtp",
  "video": "/path/124.rtp"
}
```

Steps:
- merge `audio`, `video` => `mp4 file` (basic gstreamer command)
- run `cm media-streams import-archive {streamChannelId} {mp4 file}`
- optional cleanup


fvovan:
@vogdb  could you review it please.
gstreamer pipeline is not still ready, but can you check other parts?

vogdb:
- Is `path` needed here? `bluebird` will be needed.
- If you want to use `sync` methods then don't forget to wrap them inside the Promise's body that you are going to create otherwise please use less `sync` methods.

tomaszdurka:
- yes, path not really needed (was used before)
- yes, I will use `sync` method and wrap them in the promise

fvovan:
@vogdb please review.
I'm not entirely sure where to place `.catch()`.
In addition I think it would be good `CmApplication::runCommand` returning promise as well, @tomaszdurka  wdyt?

vogdb:
- Lets use `^` or `~` in package.json for `tmp`.
- If we use `child-process-promise` then why not to use smth like this?
```js
  tmpNameAsync({postfix: '.mp4'})
    .then(function(mpegFile) {
      return exec('gst-launch-1.0', [
        'filesrc',
        'locationAudion=' + audioFile,
        '! locationVideo=' + videoFile,
        '! decodebin',
        '! videoconvert',
        '! pngenc snapshot=true',
        '! filesink location=' + mpegFile
      ]);
    })
    .then(function() {
      serviceLocator.get('cm-application').importMediaStreamArchive(streamChannelId, mpegFile)
    })
```

Have you tried it?

vogdb:
> In addition I think it would be good `CmApplication::runCommand` returning promise as well, 

+1

fvovan:
@vogdb, i was going to use `child-process-promise` but then found some confusing API in its docs (weird `.progress` and `.fail` handlers, different from BlueBird) and decides to switch back, but forgot to fix `require`

tomaszdurka:
Yes `runCommand` should return somethinglike promise, maybe also simple `spawn` will do.

I had this code on side:
```js
var exec = function(command, args, options) {
  return new Promise(function(resolve, reject) {
    var ps = require('child_process').spawn(command, args, options);
    ps.on('close', function(code) {
      if (code === ) {
        resolve();
      } else {
        reject(code);
      }
    });
  });
};
```
I think it would be nice to have simple thing like that as simple function-file - not esencially a file, wdyt?
Additionally I still consider grouping files into folders like "classes", "functional", wdyt?

vogdb:
- move `tmpNameSync` inside Promise.
- remove `catch` block. Error will be caught by `JobManager._handleJobFile`

fvovan:
>move `tmpNameSync` inside Promise.

I promisified `tmpName` instead.

vogdb:
lgtm

fvovan:
tested and debugged it with dummy shell commands
seems to work.

vogdb:
looks ok

PR #66, opened 2015-11-24
Handle plugin errors in transactions

vogdb:


vogdb:
@fvovan please review

fvovan:
lgtm, but why did you change `createResponse ` signature?

vogdb:
Because it was wrong :( I just forgot to fix it before.

PR #71, opened 2015-11-24
Fix binaryPath -> path

tomaszdurka:


tomaszdurka:
Self-merging as clear bug.

@vogdb please post-review.

PR #77, opened 2015-11-25
Expect additional data-bag

vogdb:
Expect additional data-bag for plugin events like:
- Video::onCreate
- Audio::onCreate (not there yet)

This should be sent by client using `message['data']`


vogdb:
@tomaszdurka please don't merge. It is just a draft. Do these changes describe what you meant? I don't see anything else to change for those requirements.

vogdb:
@tomaszdurka please review

njam:
Naming: how about we call those `sessionData` and `channelData`?
@tomaszdurka could you open a ticket on CM as well and discuss with Vladimir?

PR #79, opened 2015-11-25
Introduce Session class

tomaszdurka:


tomaszdurka:
image

This follows spec-like (rather than agreed before unit-like) convention 
@vogdb @fvovan wdyt?

fvovan:
But why you decided to switch to `should` DSL  style?

vogdb:
I think that committing at 4am was quite a deal for you :crescent_moon: Did you get to sleep at all?

vogdb:
- Should not `Transactions` inherit from `ObjectCollection`?
- Why `sessionId` instead of `session_id` in janus messages? Janus uses the latter.
- Why so sudden change to `expect`? It is just another flavour and also you was the one who wanted `assert` >_<

vogdb:
Ok I see why the sudden change to `expect`. To test promises in one line :)

fvovan:
I think it should not be inherited from Collections, a cause it's violation of 'single responsibility' principle. 
and I would rather use some implemented bicycle like http://www.collectionsjs.com/ for them

tomaszdurka:
@fvovan I agree, it supposed to be simply find/add, but then evolved. I wanted to add instanceof checking too. Then I thought yes, I should use some already imlemented solution - can collectionsjs do all of that?

tomaszdurka:
Closing, will be used for reference and code snippets in new pull-request.

PR #81, opened 2015-11-26
Error handling for websocket connection to janus

fvovan:
@tomaszdurka @vogdb @fvovan 

Cannot resolve host:
```
app error Uncaught exception Error: getaddrinfo ENOTFOUND localhostx
    at errnoException (dns.js:44:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:94:26)
```

Cannot connect:
```
app error Uncaught exception Error: connect ECONNREFUSED
    at exports._errnoException (util.js:746:11)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1010:19)
```

Are we handling connection errors correctly to CM?


fvovan:
@njam ahh. sorry. Actually in title you meant ws connections to janus but not to CM.
I added very simple handler for ws errors. Just logging and rethrowing. Not sure what to do during these errors. Close browser connection or something else?

njam:
Also note that this made the program exit when we tested it. Would be good if it would continue running!

vogdb:
Start to rework here. Please don't push here

vogdb:
@fvovan please look at `HttpServer.stop` and `JobManager.stop`. Can you try to implement a gentle shutdown in them?

vogdb:
Now working on tests for the new functional.

vogdb:
@fvovan can you review?

test/spec/janus-proxy.js
```diff
@@ -0,0 +1,47 @@
+var _ = require('underscore');
+var assert = require('chai').assert;
+require('../helpers/global-error-handler');
+var WebSocket = require('ws');
+var JanusProxy = require('../../lib/janus-proxy');
+var Logger = require('../../lib/logger');
+var Streams = require('../../lib/streams');
+var serviceLocator = require('../../lib/service-locator');
+
+describe.only('JanusProxy', function() {
```
vogdb: I know, I forgot this `only` again. Will fix

vogdb:
or @tomaszdurka please review?

fvovan:
looks good but probably I would use one array for 
`_connections` and `_connectionListeners`. They are always used together and could be compounded in single array of objects. But most likely is okay now.
Shouldn't we add tests for exception cases?

vogdb:
I agree that one array is better. I couldn't come up with a good name :( Only integration tests can be written for exception cases cause they are attached in `index.js` (only can be attached there)

Also I'm going to remove `stop` functionality from JanusProxy as it overcomplicates things and not necessary if we fix vendor's janus.js library

vogdb:
@fvovan please review

lib/http-server.js
```diff
@@ -97,4 +92,11 @@ HttpServer.prototype.start = function() {
   );
 };
 
+/**
+ * @returns {Promise}
+ */
+HttpServer.prototype.stop = function() {
+  return Promise.resolve();
+};
+
```
fvovan: Probably should be deleted  with its call in `index.js` as #84 is closed

lib/index.js
```diff
@@ -55,6 +55,17 @@ Application.prototype.start = function() {
 
   var jobManager = new JobManager(this.config['jobManager']['jobsPath'], []);
   jobManager.start();
+
+  process.on("unhandledRejection", function(reason) {
```
fvovan: do we have conventions to use single quotes for JS code as we have in PHP?

fvovan:
the rest looks good except it has conflicts

vogdb:
@fvovan please look

fvovan:
lgtm

PR #84, opened 2015-11-27
Gentle shutdown of http-server

fvovan:


fvovan:
@vogdb  please review. it's very basic and waits for current connections.
There is no way  (at least I didn't find them ) to gracefully finish a connection (to send some another response)  on server shutdown.
We either wait for it normal dispatching or silently terminate it

vogdb:
Please consider the error case in callback of `close`

fvovan:
it's done

vogdb:
- pass `err` to reject
- log level for minor failure should be higher than `info`. `warn`?

fvovan:
I would use `error` then

vogdb:
lgtm

fvovan:
then could be merged after https://github.com/cargomedia/cm-janus/pull/81

vogdb:
We can drop it, not crucial to have.

PR #76, opened 2015-11-27
Remove authentication for each message

tomaszdurka:
Instead of authenticating each message we will store token (json encoded data in our case) and send it only to publish/subscribe.


tomaszdurka:
@vogdb please review

vogdb:
@tomaszdurka please fix the tests

tomaszdurka:
@vogdb please re-review

vogdb:
lgtm

PR #86, opened 2015-11-30
implement graceful JobManager.stop

fvovan:


fvovan:
I created basic draft with cancellable promises.
@vogdb could you have a look? 
but the question is how to react on job handler side? Job handler can be interrupted in any place and what actually we can do? Undo archiving file with cm?

vogdb:
the draft is ok

I think that we need to delete all files of job handlers that are not finished yet. We can do it two ways:
- every job handler keeps track of all files that are created by him during its lifecycle so all of them can be deleted on shutdown.
- all job handlers create their files in some tmp directory that can be completely cleaned on shutdown. This statement is not related to job files (janus creates them) cause they must be deleted only after job handler finished successfully.

fvovan:
I would choose the latter option.
Just put all files to /tmp/cm-janus/temp-job-handler/ and completely empty it on start (if exists) and on shutdown. Btw probably would be good to attach shutdown to `process.beforeExit` event

fvovan:
@vogdb please review

fvovan:
Is it worth to write some tests for that?

vogdb:
- Lets remove confusion of names. Usually you do:

  ```js
  var audioFilename = jobData.audio;
  ```
  ```js
  var audioConvert = function(audioFilename, mpegFilename) {
  ```
  Lets inline the first usage? This one `var audioFilename = jobData.audio;`

- We should return a Promise after `handler.handle(job['data']);`

vogdb:
> Is it worth to write some tests for that?

would be great

fvovan:
>Lets inline the first usage? This one var audioFilename = jobData.audio;

Actually I am thinking  about putting these functions to class, to make them mockable for tests

lib/job/manager.js
```diff
@@ -33,7 +45,17 @@ JobManager.prototype.start = function() {
  */
 JobManager.prototype.stop = function() {
   this._removeFileListeners();
-  //TODO shutdown current working job handlers
+
+  _.each(this._activeJobs, function(jobId, jobHandle) {
```
vogdb: Lets be consistent in names. `jobHandler` here or `jobHandle` in `_handleJobFile`

vogdb:
Also please check for the case when `jobDirectory` is inside `tempFilesDirectory` and throw on it
Review done

vogdb:
not an inquiry but just a curiosity. Shall we have more narrow file permissions on `tempFilesDirectory` and `jobDirectory`. Currently we have `0777`.

fvovan:
>not an inquiry but just a curiosity. Shall we have more narrow file permissions on tempFilesDirectory and jobDirectory. Currently we have 0777.

for `tempFilesDirectory` 0777 is okay as it should be temporary  but  for `jobDirectory` it's probably  not but will add a complexity (not much certainly) to add group for janus and cm-janus. And IMO  `jobDirectory` should be located not in `/tmp` but rather in `/var` like data

vogdb:
> And IMO  `jobDirectory` should be located not in `/tmp` but rather in `/var` like data

Agree. It is a quite sensitive data to loose.

fvovan:
Tests are done.
I found one small issue with passing temp directory name to handlers. The problem is they are created outside  of `jobManager` (and even before it) and they don't know about its temporary directory. So I had to inject this attribute in `_setupJobHandlers()`. https://github.com/fvovan/cm-janus/commit/8c0be6028ac9e396f8602328c21e5fc97a8ab579 and add some duplications to prepare tempDir before it.
@vogdb could you review it please?
Regarding to changing format of config file I think it would be better to do as a separate PR and discuss there.

vogdb:
- lets use `bind` in shell methods like `AudioroomRecordingHandler.prototype._audioConvert`
- shall we place job handlers shell scripts into the same dir as job handlers files?
- move `this._tempDir = '';` inside its constructor
- lets make the relation between `1500` and `500` more obvious in test `'cleanups after shutdown'`. Create a variable for `1500` ?
- isn't `this._prepareTempFilesDir(this._tempFilesDirectory);` enough to have in constructor only? Why to have it in `start`?

fvovan:
>shall we place job handlers shell scripts into the same dir as job handlers files?

we discussed it with Tomasz. At first he proposed to store them together as well, but for me it's not good idea to mix javascript `lib` files with executable ones. Moreover these scripts should be not related  on job handlers at all. They exists independently.

>isn't this._prepareTempFilesDir(this._tempFilesDirectory); enough to have in constructor only? Why to have it in start?

You are right. It's enough in case we call `start` once. But it considers `stop` method for cleanup and to be consistent `start/stop` should do the opposite things and generally might be called unlimited number of times. I had several ideas: to move `_setupJobHandlers` to `start()`, or pass only some string keys as handlers to `jobManager` and then create handlers instances in `_setupJobHandlers`

tomaszdurka:
Guys let's stop overcomplicating this so much. We are now introducing another entry to config which is used only internal and doesn't matter in the end. Additionally we have bash scripts which totally ignore this tmp directory - unless you really want to pass it to the script too? 

- What is wrong in using global temp in general?
- What's the problem if file stays there after unexpected error? - Should NOT happen.

fvovan:
>What is wrong in using global temp in general?

`temp` is a dir that could be safely cleared after server restarting.

vogdb:
> What's the problem if file stays there after unexpected error? - Should NOT happen.

Well, all of this tries to cover the case of unexpected error. And I think that unexpected error may happen. That is why it is unexpected.

vogdb:
@tomaszdurka But we can postpone those changes if you want

fvovan:
@vogdb please review.

vogdb:
Shouldn't we move `this._tempDir = '';` to
```js
function AbstractJobHandler() {
}
```
?

The rest looks good

vogdb:
@tomaszdurka should we merge this in the end?

fvovan:
ahh sorry. forgotten to press insert

vogdb:
lgtm. lets wait for Tomasz

vogdb:
add a test where exec write to a file every 2 secs and we shutdown JobManager

vogdb:
remove streams from cm on cm-janus shutdown

fvovan:
>add a test where exec write to a file every 2 secs and we shutdown JobManager

I'm not sure what you meant, we already have similar test which writes to file but using `fs.writeFileAsync`. Which of them should be dropped?

fvovan:
>remove streams from cm on cm-janus shutdown

I've created new issue for it. https://github.com/cargomedia/cm-janus/issues/107

vogdb:
@fvovan I will adjust here some details. Please hold.

tomaszdurka:
As discussed with @vogdb deletion of the files might trigger unexpected exceptions. Keep in mind we are deleting files which are worked on by execs. Execs will be stopped only after main process is stopeed, so they might throw exception (file access problems).

@fvovan can you elaborate? In case you disagree please add test to cover this, otherwise clearing the files without earlier process shutdown will be problematic.

fvovan:
well, before it was an emergency exit when we might not care about external exec
since it has been transformed to normal 'shutdown'  I propose not clearing temp files at all, just cleanup all temp dir on application start or even delegate temp dir management to puppet

vogdb:
Please keep in mind that I'm working here and already prepared huge changes. Some minor changes are left.

vogdb:
I still don't like how JobManager tests look lie but for the time being they are ok.
@fvovan please review.
@tomaszdurka please look at the concept.

fvovan:
review is finished

vogdb:
@fvovan please review

lib/job/model/abstract.js
```diff
@@ -0,0 +1,84 @@
...
+      if (null === error) {
+        resolve();
+      } else {
+        reject(error);
+      }
+    })
+  }.bind(this));
+};
+
+AbstractJob.prototype._exec = exec;
```
fvovan: That's not needed i guess

fvovan:
the rest lgtm

lib/job/model/abstract.js
vogdb: well, we use it in tests

fvovan: oh, really, probably we shouldn't do it since scripts' execution was extracted to separate method.
as you mentioned they need to be reworked a bit later

vogdb: but then we will not be able to test this line:
```js
assert(fs.existsSync(commandArgs[0]), 'script ' + commandArgs[0] + ' does not exist');
```
is it ok?

fvovan: it's not okay certainly. 
To my mind ideally we should mock `_runJobScript()` but not its internal implementation details like `exec`. 
well we can even mock `exec` (even though i'm not sure if it's needed) but not exposing one more method only for tests usage. 
Probably it's just another matter of taste and not actually super important..

vogdb: Then lets not overcomplicate and leave as it is for now? wdyt?

fvovan: agree. let's merge it

PR #90, opened 2015-11-30
Janus http client

vogdb:


vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

lib/janus-http-client.js
```diff
@@ -0,0 +1,55 @@
+var requestPromise = require('request-promise');
+var serviceLocator = require('./service-locator');
+var ProxyConnection = require('./proxy-connection');
+
+function JanusHttpClient(url) {
+  this.baseUrl = url;
+}
+
+JanusHttpClient.prototype._httpClient = function() {
```
tomaszdurka: This method is a bit strange, same as in CmApiClient (where it was copied from?).

Why not proxy whole method aka
```
JanusHttpClient.prototype._httpRequest = requestPromise;
```
?

lib/janus-http-client.js
```diff
@@ -0,0 +1,55 @@
...
+    method: 'POST',
+    uri: this.baseUrl + action,
+    json: true
+  };
+  if (data) {
+    options.body = data;
+  }
+
+  serviceLocator.get('logger').info('janus-http-client', 'request', options.uri, options.body);
+  return this._httpClient()(options)
```
tomaszdurka: As mentioned above this is quite strange usage. We treat httpClient as method which returns method. Hmmm. Let's do as noted above or similiar.

tomaszdurka:
Rest looks ok!

lib/janus-http-client.js
fvovan: That was me, I used it to mock `request-promise`
as `requestPromise` is a function I just returned it

tomaszdurka:
Pair-finished with @fvovan

PR #83, opened 2015-12-01
"Undefined" sent to client

vogdb:
@tomaszdurka @vogdb @fvovan

```
app info cm-api request localhost { method: 'CM_Janus_RpcEndpoints.publish',
  params: 
   [ 'cm-api-key',
     'my-mountpoint',
     '0a40fcc2-f791-41f0-ba0a-28ecc062a6e4',
     1448556672278,
     undefined,
     undefined ] }
app debug -> browser undefined
```


vogdb:
Here is the fix without tests. Tests actually require to solve one problem that takes some time to deal with. What shall we do in case of publish error? What shall we do in case we have unexpected JanusError in `processMessage`? Now it leads to `unhandledRejection`. Should we close `proxyConnection` on error during `processMessage`? On some types of error, I think we should. What types then?

vogdb:
@tomaszdurka please review the principle of the new mechanism
@fvovan please do the usual review

lib/plugin/streaming.js
```diff
@@ -1,8 +1,10 @@
 var Promise = require('bluebird');
 var util = require('util');
 var PluginAbstract = require('./abstract.js');
+var ProxyConnection = require('../proxy-connection');
```
fvovan: this seems to be not used

lib/proxy-connection.js
```diff
@@ -29,19 +29,23 @@ ProxyConnection.prototype._installListeners = function() {
   this.browserConnection.on('message', function(request) {
     self.processMessage(request).then(function() {
       self.janusConnection.send(request);
-    })
+    }).catch(function(error) {
+      //error has been caught and logged by processMessage
+    });
```
fvovan: I'm not sure about node conventions and best practices, but isn't ignoring exceptions a bad practice?
It errors are caught in `processMessage` could empty `catch` block be deleted?

vogdb: Because if we don't throw then promise resolves and we pass message further to either `janusConnection` or `browserConnection` which we don't want cause an error happened. That is why we have to catch errors at this level. If we don't catch them then they will go to `unhandledRejection` listener.

vogdb:
@fvovan please review

vogdb:
@tomaszdurka please also look

lib/proxy-connection.js
```diff
@@ -175,15 +175,37 @@ ProxyConnection.prototype.stopStream = function(streamId) {
...
+
+/**
+ * @param {Error} error
+ * @returns {Promise}
  */
-ProxyConnection.prototype._processUnexpectedError = function(error) {
+ProxyConnection.prototype._unexpectedError = function(error) {
   if (!(error instanceof JanusError.Error)) {
-    serviceLocator.get('logger').warn('"' + this.identifier + '" connection processMessage error' + error);
     error = new JanusError.Unknown(ProxyConnection.generateTransactionId());
```
fvovan: I know that's not related to given PR but why to create fake transaction ids and error with them?
Could we just create a response for "Non JanusError" instances and send it instead of `error.response`

fvovan:
the rest looks good for me

lib/proxy-connection.js
vogdb: I wanted to reuse the existing mechanism that we already have for errors like `IllegalPluginError`, that's it. If you want to change it, lets do it in another PR please.

vogdb:
@fvovan can I merge here?

fvovan:
Yes sure.

PR #91, opened 2015-12-01
Rework requestPromise in CmApiClient

tomaszdurka:


tomaszdurka:
Paired with @fvovan please confirm

PR #88, opened 2015-12-01
ProxyConnection StopStream - use http instead of ws

tomaszdurka:
According to API use:
```
curl http://127.0.0.1:8300/janus/[session-id]/[plugin-id] -H "Content-Type: application/json" -X POST -d '{"janus": "detach", "transaction": "[transaction-id]"}'
```

Store session-id and plugin-id on Stream instance only (no need for instances).


tomaszdurka:
@vogdb please review

vogdb:
Looks very good. Two notes:
- Unused dependency in `test/spec/http-server.js` is underscore
- One feature that I forgot to mention before in `janus-http-client`. `stopStream` can receive an `ack` response instead of immediate `success` response. That means your request is acknowledged and stream is not actually stopped. Should we consider this situation in our client?

vogdb:
And I think you see that tests failed.

vogdb:
@tomaszdurka ping me if you need help here

tomaszdurka:
@vogdb I agree it might be important to be sure stream has really stopped. Btw maybe we should even detach whole plugin. Let's move this into new issue and investigate soon. I would be up for merging this and working on structuring janus resources bit more.

vogdb:
lgtm

PR #93, opened 2015-12-02
Fix: properly cleanup websocket connection between cm-janus and janus

vogdb:
My test shows that if websocket connection is closed between any `client` (studio-agent) and the `cm-janus` then the connection between `cm-janus` and `janus` is still alive?

could you verify?


vogdb:
@tomaszdurka please review

vogdb:
oh, nobody is online. @kris-lab please review

kris-lab:
lvgtm!

PR #97, opened 2015-12-03
Introduce janus/session

tomaszdurka:


tomaszdurka:
@vogdb please review

vogdb:
started

vogdb:
- `var pluginId = message['handle_id'] || message['sender'] || null` Why to use `null` in the end?
- `ProxyConnection.prototype.onAttach` still exists and contains more fresh and correct version of `onAttach`
- Redundant `var _ = require('underscore');` in ProxyConnection
- `proxyConnection.sessionId` is still used. `http-server` for example
- What is the point of single usage of `libRequire` ?

Now reviewing tests

vogdb:
- Here is an example of `destroy` message
```json
{
  "janus":"destroy",
  "transaction":"qhxQSDLbsUSi",
  "token":"123fish",
  "session_id":1074764448
}
```
  I think that we should use the same format in our tests and also do not handle `destroy` message if it does not have a `session_id` field. Test `'when processes "destroy" message"'` uses `sessionId` instead of `session_id` and `processMessage` of `ProxyConnection` allows to destroy without correct `session_id`.

vogdb:
- empty test `it('should clear transactions');`

Review is done

tomaszdurka:
Code:
- This is to pass `null` instead of `undefined`. Actually it was like that before (?).
- Bad merge of upstream. Resolved.
- Resolved.
- Resolved.
- This was just simple addition to show how it works. Not sure if it's good idea to put it everywhere (maybe only in tests?) - so wanted to know your opinion.

Tests:
- I left empty as it's not really supported, but should it be implemented at all? I guess it wouldn't be right - removed.

tomaszdurka:
I know coverage dropped, especially in proxy-connection, but I will improve it in #98.

tomaszdurka:
@vogdb please review

vogdb:
- `var sessionId = stream.plugin.proxyConnection.sessionId;` in `stopStream` of HttpServer.
- `libRequire`. lets not use it.
- `onDestroy` must belong to Session cause it contains `session_id` and will be handled in `processMessage` of Session

vogdb:
review is done

tomaszdurka:
Done
- didn't find sessionId anymore
- libRequire removed
- `onDestroy` needs to be within connection to remove the session, same for attach/detach plugin to remove plugin. I have already changed this in #98. Here now too.

vogdb:
@tomaszdurka please resolve conflicts

tomaszdurka:
Done

vogdb:
lgtm!

PR #98, opened 2015-12-03
Rework proxy and connection, session, plugin resources

tomaszdurka:


tomaszdurka:
@vogdb please review.

This is probably the pre-last set of changes I wanted to make. 
This change-set includes:

After testing the cm-janus manually i felt we have couple of problems:
- possibly not clearing streams properly on certain events (e.g. plugin detach)
- possible memory leaks (e.g. because above)
- bit spread error handling
- spreading responsibility for actions and their callbacks (effects)


This pull-request continues to:
- introduce simple (possibly non-communicational) resources (Connection, Session, Plugin)
- above resources have 2 main purposes - store data and interpret incoming/outgoing messages
- cascade removal of resources (if connections closes it removes sessions, plugins and resources of plugin - stream in future; etc).

In next step (pull-request):
- move storing of streams into plugins
- collect stream status data by iterating through connections, sessions and their plugins (single source of truth)
- remove streams collection
- remove streams by and only by removing it from plugin
- possibly stop stream by detaching plugin instead of stopping stream
  - this is to properly notify the websocket connection about the change
  - currently `stop` event sent via http api will not trigger any event on existing websocket connection

tomaszdurka:
Again spec coverage is dropping this will be heavily increased in last pull-request.
THANKS!

vogdb:
Started

vogdb:
I see massive PR attack that depends on this. So, we can merge here.

PR #92, opened 2015-12-04
Fix chanelData read in streaming pluign

tomaszdurka:
Should use `message['body']['channel_data']` instead of `message['channelData']`.


tomaszdurka:
@kris-lab please review

kris-lab:
lgtm

PR #96, opened 2015-12-04
Datetime is passed in miliseconds in create/webrtcup event

tomaszdurka:
Should be seconds.


tomaszdurka:
@kris-lab please review

kris-lab:
looks good

PR #100, opened 2015-12-04
Move Janus classes into janus/ directory

tomaszdurka:

PR #101, opened 2015-12-04
Rework streams persistence

tomaszdurka:


tomaszdurka:
Depends on #98, #99, #100.

tomaszdurka:
@vogdb please review

vogdb:
- why do we duplicate this test in `spec/janus/connection`? Increase the coverage?
```js
      it('should clear session', function() {
        connection.onRemove();
        expect(connection.session).to.be.equal(null);
      });
```
- unused `var Promise = require('bluebird');` in `spec/janus/plugin/abstract`
- unused `var JanusConnection = require('./janus/connection');`, `var serviceLocator = require('./service-locator');` in `lib/streams.js`
- unused `var PluginAbstract = require('../../../../lib/janus/plugin/abstract');`, `` in `spec/janus/plugin/streaming.js`
-why not to use single style among `serviceLocator.register('logger', sinon.stub(new Logger));` and `serviceLocator.register('cm-api-client', sinon.createStubInstance(CmApiClient));`. Is there a difference between them?
- shouldn't we use `executeTransactionCallback().then` instead of `executeTransactionCallback().finally` in tests `'on successful transaction response'`? Why `finally` is used?
- Are these tests complete?
```js
      context('on successful publish', function() {
        it('should add stream to streams', function(done) {
          executeTransactionCallback().finally(function() {
            done();
          });
        });
      });
```
```js
it('should detach');
```
```js
    context('on successful subscribe', function() {
      it('should add stream to streams', function(done) {
        processWebrtcupMessage().finally(function() {
          done();
        });
      });
    });
```
- unused `var _ = require('underscore');` in `spec/janus/plugin/audio.js`

tomaszdurka:
- (1) I believe it was copied to be base for another test. Removed.
- (2) Removed unused dependencies.
- (3) Removed unused dependencies. Also removed event-emitter as it's not used anywhere anymore.
- (4) Removed unused dependencies.
- (5) Yes, there is a difference. I prefer using `sinon.createStubInstance()`, but it only touches the prototype of the class. While `Logger` adds `debug`, `info`, etc. methods in constructor. That's why I use `stub` (which can be used on object) after constructor. Bit hacky. Probably best solution is to add logger methods into prototype.
- (6) Well this is a bit tricky. So `on successful transaction response` means that there is a success-message coming from janus. It doesn't mean yet if the transaction callback execution will succeed. I used finally because stream should be set, and it should publish no matter of the transactionCallback (as whole) result.  Because depending on `publish` return (resolved/rejected) we will determine the transactionCallback() execution. I thought about this test for long time. I have split this into several methods for a moment (with separate spec for `publish`), but then went back to current state.
- (7) These tests are not complete. I was to add them, forgot. In case of `detach` we still miss functionality to do this (http call).
- (8) Removed unused dependencies.

tomaszdurka:
@vogdb please review

vogdb:
Short notes:

 - 'spec/janus/proxy' unused `var JanusConnection = require('../../../lib/janus/connection');`
 - In `janus/connection.js`. What was the use case for?

  ```js
  if (this.session.plugins[pluginId]) {
    return this.session.processMessage(message);
  }
  ```
 - Why do we still keep `JanusHttpClient.prototype.stopStream` if we don't use it.

tomaszdurka:
- ok.
- detached :/ there was no `session_id` was ignored - not passed to session :/
- i wanted later (next pr) while I add tests for detach at the same time.

vogdb:
ok, lgtm then

PR #104, opened 2015-12-11
RPC Error logging

vogdb:
When an RPC call to the CM application fails, the application responds with a JSON structure of this form:
```json
{"type": "CM_Exception_AuthRequired", "msg": "Internal Server Error", "isPublic": false}
```

Afaik cm-janus currently only displays the `msg`. Logging additionally the `type` would be handy.
@tomaszdurka @fvovan @vogdb wdyt?

### DIFF
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -41,10 +41,12 @@ CMApiClient.prototype._request = function(action, data) {
       throw new Error('cm-api error: ' + error['message']);
     })
     .then(function(response) {
-      serviceLocator.get('logger').info('cm-api', 'response', response.body);
-      if (response['error']) {
-        throw new Error('cm-api error: ' + response['error']['msg']);
+      var error = response['error'];
+      if (error) {
+        serviceLocator.get('logger').error('cm-api', 'response', response);
+        throw new Error('cm-api error: ' + error['type'] + '. ' + error['msg']);
       }
+      serviceLocator.get('logger').info('cm-api', 'response', response.body);
       return response['success']['result'];
     });
 };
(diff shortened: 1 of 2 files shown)


njam:
I think it's okay to send the HTML in that case.
But in case of 200 it would be nice if we would also forward `type`. Should we concatenate them?

lib/cm-api-client.js
```diff
@@ -41,10 +41,12 @@ CMApiClient.prototype._request = function(action, data) {
       throw new Error('cm-api error: ' + error['message']);
     })
     .then(function(response) {
-      serviceLocator.get('logger').info('cm-api', 'response', response.body);
-      if (response['error']) {
-        throw new Error('cm-api error: ' + response['error']['msg']);
+      var error = response['error'];
+      if (error) {
+        serviceLocator.get('logger').error('cm-api', 'response', response);
```
tomaszdurka: Btw I don't think this should be logged as error.

vogdb: ok but the original `serviceLocator.get('logger').info('cm-api', 'response', response.body);` does not work correctly too cause the error response does not have `body` field which makes `response.body` to be `undefined`.
Considering the fact that we still throw an error with 200 JSON error response, I think that we should log it as an error.
What are your alternatives? @tomaszdurka

tomaszdurka: Ah, ok so put my 5 cents wrong. I thought this is about `error` property of response (body) already. I guess this is an error then.

vogdb:
@tomaszdurka please review

tomaszdurka:
Looks ok

PR #106, opened 2015-12-15
rpc_removeStream never called

vogdb:
During testing we experienced that while the `publish()` and `subscribe()` endpoints are called it seems like `rpc_removeStream()` is not!

Could reproduce this when closing WebSocket connection to cm-janus.

@tomaszdurka can you look into this and/or reassign?

### DIFF
--- a/lib/janus/plugin/streaming.js
+++ b/lib/janus/plugin/streaming.js
@@ -75,6 +75,7 @@ PluginStreaming.prototype.onWebrtcup = function(message) {
 
 PluginStreaming.prototype.removeStream = function() {
   if (this.stream) {
+    serviceLocator.get('cm-api-client').removeStream(this.stream.channelName, this.stream.id);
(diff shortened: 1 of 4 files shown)


vogdb:
@tomaszdurka please review

vogdb:
You can see a problem here is that we don't wait for response of `cmApiClient.removeStream` but I don't see an easy way to do it. We need to change the whole `onRemove` logic and make it async.

tomaszdurka:
Hmmm, true. In case we have async operation here it could/should be the case. Let's merge as is.

tomaszdurka:
On the other hand there is no much we can do if this request fail. We will log it as unhandled rejectection anyway.

vogdb:
ok, maestro!

vogdb:
@njam please try cm-janus 0.1.2

njam:
:+1: works fine in the first test

PR #107, opened 2015-12-16
Remove streams from cm on cm-janus shutdown

fvovan:


fvovan:
@vogdb  could you review it please? 
not sure where to place streams cleanup code. `index.js` seems to be the most appropriate one.

tomaszdurka:
By adding method to remove streams on index.js we are adding kinda global code - stream-specific code into entry point.

I am not sure if streams are the only ones we should remove. Maybe some plugins have other things which should be removed.
What about adding `stop` method into `janus-proxy` which would `remove` all conenctions? 
- connections would remove sessions
- sessions would remove plugins
- plugins would remove streams
We would be sure we cleaned up everything as we usually do, no need to add another usage of cm-http-client.

In general I think index.js should be global *entry point* which starts sub-apps (jobhandler, proxy, api-server) and stops all of them on incoming signal. I don't think it should know anything about streams in the end... except it needs to register service like that :/

fvovan:
>In general I think index.js should be global entry point which starts sub-apps (jobhandler, proxy, api-server) and stops all of them on incoming signal. I don't think it should know anything about streams in the end... except it needs to register service like that :/

yes, i completely agree. just didn't know where to put it.
now it's not even needed, streams will be removed within plugins' deletion.
@tomaszdurka please take a look

lib/index.js
```diff
@@ -62,7 +63,16 @@ Application.prototype.start = function() {
 
   process.on('uncaughtException', function(err) {
     serviceLocator.get('logger').error('Unexpected runtime error. Shutdown the process ' + err.stack || err.message);
-    jobManager.stop().finally(function() {
+    process.emit('close');
+  });
+
+  process.on('SIGINT', function() {
```
tomaszdurka: Listen for SIGTERM too.

tomaszdurka:
As in in-code comment - should we listen for SIGTERM?

Rest lgtm.

PR #108, opened 2015-12-17
Zombie audio publishers in status?

vogdb:
After some testing, where usually a `rpc_subscribe()` call would disallow subscribing, we ended up with lots of "zombie" audio publishers. Those were not connected anymore to cm-janus. Restarting janus and cm-janus removed them through syncing.
screen shot 2015-12-16 at 18 06 13

@tomaszdurka how does this status work? Is it proxied to janus, or it keeps its own state?
What could cause this?


vogdb:
@tomaszdurka please look. Also I think that it would be more correct to create stream for audio plugin only after `configure event`. Like this:
```js
/**
 * @param {Object} message
 * @returns {Promise}
 */
PluginAudio.prototype.processMessage = function(message) {
  var janusMessage = message['janus'];

  if ('message' === janusMessage && 'configure' === message['body']['request']) {
    return this.onConfigure(message);
  }

  return PluginAudio.super_.prototype.processMessage.call(this, message);
};

/**
 * @param {Object} message
 * @returns {Promise}
 */
PluginAudio.prototype.onConfigure = function(message) {
  if (message['jsep'] && 'offer' == message['jsep']['type']) {
    var plugin = this;
    this.session.connection.transactions.add(message['transaction'], function(response) {
      if (plugin._isSuccessResponse(response)) {
        plugin.stream = Stream.generate(response['plugindata']['data']['id'], plugin);
        serviceLocator.get('logger').info('Added ' + plugin.stream + ' for ' + plugin);
      }
      return Promise.resolve(response);
    });
  }
  return Promise.resolve(message);
};
```

vogdb:
As discussed we only subscribe in audio on webrtcup and we create stream on Join.

vogdb:
It seems that I already implemented in the way that we want. @tomaszdurka please review.

tomaszdurka:
Ah true. So we just removed the publish from create.

vogdb:
@tomaszdurka so what is your final resolution here?

tomaszdurka:
lgtm

tomaszdurka:
But remember this depends on https://github.com/cargomedia/cm/issues/2029, at least before releasing it.

tomaszdurka:
@vogdb let's add #115 here.

tomaszdurka:
@vogdb please review last commit.

@zazabe please adjust all `watch`, `join`, `switch` requests to include `['body']['channel_data']`.

vogdb:
lgtm

vogdb:
hmm cargomedia/cm#2029 wasn't yet merged.

tomaszdurka:
Yes, this still needs a release - I thnk we can set it up all now.

PR #110, opened 2015-12-18
Support switch event for video plugin

vogdb:
https://github.com/cargomedia/sk/blob/master/library/VR/client-vendor/after-body-source/vr/janus/plugin/video.js

Intercept `switch` event:
- remove previous stream (and store new one) on incoming client message
- subscribe new stream on janus response

### DIFF
--- a/lib/janus/plugin/video.js
+++ b/lib/janus/plugin/video.js
@@ -23,6 +23,10 @@ PluginVideo.prototype.processMessage = function(message) {
     return this.onWatch(message);
   }
 
+  if ('message' === janusMessage && 'switch' === message['body']['request']) {
+    return this.onSwitch(message);
+  }
+
   return PluginVideo.super_.prototype.processMessage.call(this, message);
 };
 
@@ -36,6 +40,24 @@ PluginVideo.prototype.onWatch = function(message) {
     if (plugin._isSuccessResponse(response)) {
       plugin.stream = Stream.generate(message['body']['id'], plugin);
       serviceLocator.get('logger').info('Added ' + plugin.stream + ' for ' + plugin);
+    }
+    return Promise.resolve(response);
+  });
+  return Promise.resolve(message);
+};
+
+/**
+ * @param {Object} message
+ * @returns {Promise}
+ */
+PluginVideo.prototype.onSwitch = function(message) {
+  var plugin = this;
+  plugin.removeStream();
+  plugin.stream = Stream.generate(message['body']['id'], plugin);
+  serviceLocator.get('logger').info('Added ' + plugin.stream + ' for ' + plugin);
+  plugin.session.connection.transactions.add(message['transaction'], function(response) {
+    if (plugin._isSuccessResponse(response)) {
+      return plugin.onWebrtcup(response);
     }
     return Promise.resolve(response);
   });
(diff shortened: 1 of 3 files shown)


vogdb:
@tomaszdurka I'm not sure that we want to reject on `failure transaction response` cause client does not receive the original error and can't proceed accordingly. Please lets discuss this topic.

tomaszdurka:
I am not sure if i understand. All i've described is on cm-janus level.
Client would send switch request and get one of these:
- success event proxied from janus (in case of successful janus and successful subscribe)
- error event proxied from janus (in case of janus error, no subscribe called)
- error event generated by cm-janus (in case of janus success, but subscribe error)

vogdb:
That is how it's done now. I thought that you want to generate error in this case too:

> - error event proxied from janus (in case of janus error, no subscribe called)

@tomaszdurka please review

lib/janus/plugin/video.js
```diff
@@ -42,4 +46,23 @@ PluginVideo.prototype.onWatch = function(message) {
...
+ * @returns {Promise}
+ */
+PluginVideo.prototype.onSwitch = function(message) {
+  var plugin = this;
+  plugin.removeStream();
+  plugin.stream = Stream.generate(message['body']['id'], plugin);
+  serviceLocator.get('logger').info('Added ' + plugin.stream + ' for ' + plugin);
+  plugin.session.connection.transactions.add(message['transaction'], function(response) {
+    if (plugin._isSuccessResponse(response)) {
+      return plugin.onWebrtcup(response);
```
tomaszdurka: Let's extract `subscribe` method out of `onWebrtcup` and use it here.

tomaszdurka:
Currently it doesn't remove the stream from streams/cm-app.
I tried to write down the spec in earlier comment to cover all the cases. You can freely use/implement it :)

vogdb:
> Currently it doesn't remove the stream from streams/cm-app.
I tried to write down the spec in earlier comment to cover all the cases. You can freely use/implement it :)

I thought that `plugin.removeStream` removes stream from streams/cm-app. Am I wrong? Or did you mean that we shouldn't remove it currently?
```js
PluginVideo.prototype.onSwitch = function(message) {
  var plugin = this;
  plugin.removeStream();
  ...
```

vogdb:
@tomaszdurka please review

test/spec/janus/plugin/video.js
```diff
@@ -90,4 +96,82 @@ describe('Video plugin', function() {
...
+    };
+    var switchResponse = {
+      janus: 'event',
+      plugindata: {data: {streaming: 'event', result: {}}},
+      sender: plugin.id,
+      transaction: switchRequest.transaction
+    };
+
+    plugin.processMessage(switchRequest).then(function() {
+      connection.transactions.execute(switchRequest.transaction, switchResponse).then(function() {
```
tomaszdurka: Please check that previous stream was removed.

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka Have you seen how this change decreased the test coverage?))

tomaszdurka:
Looks ok.

PR #112, opened 2015-12-21
Error message should refer to a transaction

vogdb:
The vendor janus client [only call `onmessage` when the `transaction` id is set](https://github.com/cargomedia/sk/blob/master/library/VR/client-vendor/after-body/04-janus/janus.js#L326), but cm-janus specific error have no reference to a transaction, for example:

```
{"janus":"error","error":{"code":490,"reason":"Cannot subscribe: Stream{\"id\":\"049626f1-7909-4e52-af74-d60c00f41e95\"} error: cm-api error: CM_Exception_NotAllowed. Internal server error"}}
{"janus":"error","transaction":"9KuoI217vK2O","error":{"code":490,"reason":"Invalid plugin"}}
```

2 solutions:
- cm-janus send a valid message, with the transaction id
- change the vendor client source code to call the `janus.error` callback if there's no transaction id

related to cargomedia/sk#2507

@njam fyi

### DIFF
--- a/lib/janus/plugin/streaming.js
+++ b/lib/janus/plugin/streaming.js
@@ -69,7 +69,8 @@ PluginStreaming.prototype.subscribe = function(message) {
     })
     .catch(function(error) {
       serviceLocator.get('http-client').detach(plugin);
-      throw new JanusError.Error('Cannot subscribe: ' + plugin.stream + ' error: ' + error.message, 490, message['transaction']);
+      var transaction = message['transaction'] || require('../connection').generateTransactionId();
+      throw new JanusError.Error('Cannot subscribe: ' + plugin.stream + ' error: ' + error.message, 490, transaction);
     });
 };
 
--- a/lib/janus/plugin/streaming.js
+++ b/lib/janus/plugin/streaming.js
@@ -69,8 +69,7 @@ PluginStreaming.prototype.subscribe = function(message) {
     })
     .catch(function(error) {
       serviceLocator.get('http-client').detach(plugin);
-      var transaction = message['transaction'] || require('../connection').generateTransactionId();
-      throw new JanusError.Error('Cannot subscribe: ' + plugin.stream + ' error: ' + error.message, 490, transaction);
+      throw new JanusError.Error('Cannot subscribe: ' + plugin.stream + ' error: ' + error.message, 490, message['transaction']);
     });
 };
 
--- a/lib/janus/error.js
+++ b/lib/janus/error.js
@@ -22,14 +22,17 @@ function JanusError(message, code, transaction) {
 util.inherits(JanusError, Error);
 
 JanusError.prototype.getWebSocketMessage = function() {
-  return {
+  var message = {
     janus: 'error',
-    transaction: this.transaction,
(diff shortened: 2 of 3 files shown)


vogdb:
@fvovan please review

lib/janus/plugin/streaming.js
```diff
@@ -69,7 +69,8 @@ PluginStreaming.prototype.subscribe = function(message) {
     })
     .catch(function(error) {
       serviceLocator.get('http-client').detach(plugin);
-      throw new JanusError.Error('Cannot subscribe: ' + plugin.stream + ' error: ' + error.message, 490, message['transaction']);
+      var transaction = message['transaction'] || require('../connection').generateTransactionId();
```
fvovan: Can random (I mean not related to real transaction) transaction id be useful in error messages? may be use empty string when it's absent?

zazabe: @fvovan just be careful with empty transaction id, because we will not be able to catch these errors on the client side in this case:
```
} else if(json["janus"] === "error") {
	// Oops, something wrong happened
	Janus.error("Ooops: " + json["error"].code + " " + json["error"].reason);	// FIXME
	var transaction = json["transaction"];
	if(transaction !== null && transaction !== undefined) {
		var reportSuccess = transactions[transaction];
		if(reportSuccess !== null && reportSuccess !== undefined) {
			reportSuccess(json);
		}
		delete transactions[transaction];
	}
	return;
```

https://github.com/cargomedia/sk/blob/master/library/VR/client-vendor/after-body/04-janus/janus.js#L330

vogdb: Unfortunately we can't reach any other transaction except a random one. @zazabe will it be ok for `janus.js` client if the transaction was not generated on its side?

fvovan: will it work with some predefined value like '0'?

zazabe: I think you should add the transaction only if you have a real one.
This piece of code is made to call the plugin `onmessage` callback, but the client is only aware of transaction made by him or it will just not handle the error at all... 
I already added a fix to call the janus error callback if there's no transaction id, which is fine for me...

tomaszdurka: @vogdb why there is no transaction id? I would not generate random id here. Errors are responses to some exisitng transactions. Don't think there is a reason to send one without request.

@zazabe I'm quite confused what was the original problem here. I thought that there will be no error-callback called when transaction id is not present. From the code seems valid transactionId is needed for calling `reportSuccess`, otherwise we will call global error handler. I understand that `reportSuccess` is custom transaction callback set when you request.
Seems that adding random transaction id won't change anything as it will look for callback in transactions hash.
Still. Do we need this? I don't see a problem with no transaction id.

vogdb:
@zazabe this change makes possible to send error janus messages with no transaction. Is that fine?
@fvovan please review

fvovan:
lgtm

zazabe:
@vogdb yes, only set the `transaction` if you have it is fine.

PR #114, opened 2015-12-28
Introduce acl for audio plugin

vogdb:
Allow for all:
- create 
- join (detach if not able to subscribe)
- changeroom (detach if not able to subscribe)
- configure
- leave

Deny for all:
- exists
- resetdecoder
- listparticipants
- destroy
- list

### DIFF
--- a/lib/janus/plugin/abstract.js
+++ b/lib/janus/plugin/abstract.js
@@ -18,7 +18,7 @@ function PluginAbstract(id, type, session) {
  * @returns {Promise}
  */
 PluginAbstract.prototype.processMessage = function(message) {
-  if (this.isValidMessage(message)) {
+  if (this.isAllowedMessage(message)) {
     return Promise.resolve(message);
   } else {
     return Promise.reject(message);
@@ -28,7 +28,7 @@ PluginAbstract.prototype.processMessage = function(message) {
 /**
  * @param {Object} message
  */
-PluginAbstract.prototype.isValidMessage = function(message) {
+PluginAbstract.prototype.isAllowedMessage = function(message) {
   return !message['janus'] || 'destroy' != message['janus'];
 };
 
--- a/lib/janus/plugin/abstract.js
+++ b/lib/janus/plugin/abstract.js
@@ -18,18 +18,18 @@ function PluginAbstract(id, type, session) {
  * @returns {Promise}
  */
 PluginAbstract.prototype.processMessage = function(message) {
-  if (this.isAllowedMessage(message)) {
-    return Promise.resolve(message);
-  } else {
+  if (!this.isAllowedMessage(message)) {
     return Promise.reject(message);
   }
+  return Promise.resolve(message);
 };
 
 /**
  * @param {Object} message
  */
 PluginAbstract.prototype.isAllowedMessage = function(message) {
-  return !message['janus'] || 'destroy' != message['janus'];
+  var isDisallowed = message['janus'] && 'destroy' === message['janus'];
+  return !isDisallowed;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


vogdb:
@tomaszdurka please review

lib/janus/plugin/abstract.js
```diff
@@ -17,7 +18,18 @@ function PluginAbstract(id, type, session) {
...
+  } else {
+    return Promise.reject(message);
+  }
+};
+
+/**
+ * @param {Object} message
+ */
+PluginAbstract.prototype.isValidMessage = function(message) {
+  return !message['janus'] || 'destroy' != message['janus'];
```
tomaszdurka: Is this actually right? 
Now when there is no 'janus' property it will be valid?

```js
return !undefined || 'whatever'; // -> true
```
Right?

Maybe rather return early when no `janus`, etc?

vogdb: Well, the message will be valid if there is no 'janus' field or it is present but not equal 'destroy'. I actually rely here on lazy calculation of boolean expressions.

vogdb: @tomaszdurka will that be fine?

```js
PluginAbstract.prototype.isValidMessage = function(message) {
  if (!message['janus']) {
    return true;
  }
  return 'destroy' != message['janus'];
};
```

vogdb: @tomaszdurka ?

tomaszdurka: Ok so isValidMessage has completely different meaning than I understood.
Maybe naming is a bit misleading.
isValidMessage to me should return true if message is proper janus message with janus field and one of the valid janus fields.

But this is more like `isAllowed` message?

vogdb: well, yes

tomaszdurka: For clarity and return-early rule let's change couple of small things:
- rename method to `isAllowedMessage`
- (more of preference) rather do:
```js
if (!this.isAllowedMessage(message)) {
  return Promise.reject(message);
}
// previous code below
return Promise.resolve(message);
```
- in the `isAllowedMessage` methods do not make fancy booleans but rather explicitly do:
```js
PluginAbstract.prototype.isAllowedMessage = function(message) {
  if (message['janus'] && 'destroy' === message['janus']) {
    return false;
  }
  return true;
}
```
or introduce variable?
```js
PluginAbstract.prototype.isAllowedMessage = function(message) {
  var isDisallowed = message['janus'] && 'destroy' === message['janus']);
  return !isDisallowed;
}
```

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please look

njam:
@tomaszdurka status?

vogdb:
@tomaszdurka ?

lib/janus/plugin/audio.js
```diff
@@ -26,6 +26,14 @@ PluginAudio.prototype.processMessage = function(message) {
...
 };
 
+PluginAudio.prototype.isAllowedMessage = function(message) {
+  if (PluginAudio.super_.prototype.isAllowedMessage(message)) {
+    var isDisallowed = 'message' === message['janus'] && _.contains(['list', 'exists', 'resetdecoder', 'listparticipants'], message['body']['request']);
+    return !isDisallowed;
+  } else {
+    return false;
+  }
+};
```
tomaszdurka: Looks good, maybe for same clarity rule you could:
```
if (!PluginAudio.super_.prototype.isAllowedMessage(message)) {
  return false;
}
// additional rules
```
But is ok for me.

tomaszdurka:
Looks good!

Issue #109, opened 2015-12-17
Failed to stop a video stream


zazabe:
Not sure if it's an issue on janus or cm-janus, but there's definitely something wrong with the `rtpbroadcast` plugin.

cm-janus log when the client do a `"request": "stop"`:

```
app debug <- browser {"janus":"message","body":{"request":"stop"},"transaction":"jNTKBusJp4BP","token":"{\"sessionId\":\"feeb77fdd859672948b28c411ffa72a3\"}","session_id":712322919,"handle_id":3045102479}
app debug -> janus {"janus":"message","body":{"request":"stop"},"transaction":"jNTKBusJp4BP","token":"{\"sessionId\":\"feeb77fdd859672948b28c411ffa72a3\"}","session_id":712322919,"handle_id":3045102479}
app debug <- janus {"janus":"ack","session_id":712322919,"transaction":"jNTKBusJp4BP"}
app debug <- janus {"janus":"hangup","session_id":712322919,"sender":3045102479,"reason":"DTLS alert"}
app info cm-api request https://alicex.dev.cargomedia.ch/rpc/null {"method":"CM_Janus_RpcEndpoints.removeStream","params":["mad-panda","mD9ieC9oATnpHVo8HcyjYg__","b9b53ff9-6c31-4ed4-80d5-89a5ce5a5319"]}
app info Removed Plugin{"id":3045102479,"type":"janus.plugin.cm.rtpbroadcast"}
app debug -> browser {"janus":"ack","session_id":712322919,"transaction":"jNTKBusJp4BP"}
app debug -> browser {"janus":"hangup","session_id":712322919,"sender":3045102479,"reason":"DTLS alert"}
app info cm-api response
app debug <- janus {"janus":"event","session_id":712322919,"sender":3045102479,"transaction":"jNTKBusJp4BP","plugindata":{"plugin":"janus.plugin.cm.rtpbroadcast","data":{"streaming":"event","result":{"status":"stopping"}}}}
app info {"transaction":"jNTKBusJp4BP","code":490,"message":"Invalid plugin"}
app debug -> browser {"janus":"error","transaction":"jNTKBusJp4BP","error":{"code":490,"reason":"Invalid plugin"}}
```

We should not have the last janus error response to the browser i suppose...

kris-lab:
[...]

PR #116, opened 2015-12-30
Do not trigger errors on nonexistent plugins/sessions

tomaszdurka:


tomaszdurka:
@vogdb please review

Fixes https://github.com/cargomedia/cm-janus/issues/109

vogdb:
This is a huge security hole IMO cause client can send a message for session/plugin which he does not own and do whatever he wants. But I hope that I'm wrong :)

lgtm

tomaszdurka:
Let's go with this solution for now @vogdb and let janus deal with it.

vogdb:
ok ok

vogdb:
you can merge

PR #125, opened 2016-01-10
Introduce Channel class

tomaszdurka:


tomaszdurka:
Depends on #124

tomaszdurka:
@vogdb please review

vogdb:
- tests for Channels.
- Why not 

  ```js
  Channels.prototype.add = function(channel) {
    if (this.contains(channel)) {
      throw new Error('Channel with name `' + channel.name + '` already exists');
    }
    this.list[key] = channel;
  };
  ```
- `spec/streaming.js`. Why recreate `plugin.stream` with a new stream?

  ```js
      var channel = {name: 'channel-name', data: 'channel-data'};
      var stream = new Stream('stream-id', channel);
      plugin.stream = new Stream('stream-id', 'channel-name', 'channel-data', plugin);
  ```

tomaszdurka:
@vogdb 

- Will add
- I check that either channel name or key is there.
- That's mistake

vogdb:
> - I check that either channel name or key is there.

Why to concatenate them then for unique key? If any of them is unique then we can just use `id` for example.

tomaszdurka:
@vogdb please review

vogdb:
- Channels.prototype.findByName can be simplefied through `this.list[channel.name]`?
 - Does it make sense to have `PluginStreaming.prototype.onCreate` if we have `PluginAudio.prototype.onCreate` that is independent of its parent?
 - Should we register `serviceLocator.register('channels', channels)` in `index.js`?

tomaszdurka:
- Let's keep it as is. Maybe key will change from name one day, it's bit more clear and almost same performant.
- Oh there is some bug here. I think it was to be moved to `VideoPlugin` and we moved without changing the name of class :panda_face: Ah, but it's only processed in Video plugin. Messy/hacky/bug. In the end we should have 2 separate onCreate handlers for Video as is, for Audio without publish part.
- Oh yeah! :)

tomaszdurka:
@vogdb please review

vogdb:
Hoooray! lgtm

PR #127, opened 2016-01-12
Verify if cleanup is done when "mountpoint" is destroyed

tomaszdurka:
@tomaszdurka @vogdb @njam

We should make sure that `cm-janus` is aware of `mountpoint` destroy event. If `mountpoint` is destroyed of any reason e.g. master session expires then Janus removes `mountpoint` and notify all connected sessions to the `rtpbroadcast` plugin. It sends following messages:

```
{
   "janus": "event",
   "session_id": 1734302109,
   "sender": 2805623492,
   "plugindata": {
      "plugin": "janus.plugin.cm.rtpbroadcast",
      "data": {
         "streaming": "event",
         "result": {
            "status": "stopped"
         }
      }
   }
}
```
and
```
{
   "janus": "hangup",
   "session_id": 1734302109,
   "sender": 2805623492,
   "reason": "DTLS alert"
}
```

We should make sure to remove `stream` in `CM` then with e.g. `rpc_removeStream`?

@tomaszdurka can you confirm it is implemented?

### DIFF
--- a/lib/janus/plugin/video.js
+++ b/lib/janus/plugin/video.js
@@ -33,7 +33,7 @@ PluginVideo.prototype.processMessage = function(message) {
     return this.onSwitch(message);
   }
 
-  if ('event' === janusMessage && 'streaming' === message['plugindata']['data']['streaming'] && 'stopped' === message['plugindata']['data']['result']['status']) {
+  if ('event' === janusMessage && 'event' === message['plugindata']['data']['streaming'] && 'stopped' === message['plugindata']['data']['result']['status']) {
     return this.onStopped(message);
   }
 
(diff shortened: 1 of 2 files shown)


tomaszdurka:
@vogdb please review

tomaszdurka:
@vogdb you can also adjust this to your newer switch-case style.

vogdb:
I would use `_.isMatch` instead of `_.isEqual` and also added some tests. Do you want me to do it?

vogdb:
Seems lgtm. @tomaszdurka ?

vogdb:
@tomaszdurka please review

tomaszdurka:
Looks good!

PR #153, opened 2016-01-17
Introduce job working directory

tomaszdurka:


tomaszdurka:
@vogdb please review

tomaszdurka:
@vogdb please review

Changes here are mostly to create files and run scripts in single working directory which is:
- created before
- removed after

This way we don't need to care about individual files and problems with removing them in scripts.

vogdb:
@tomaszdurka lgtm, please verify

Issue #142, opened 2016-01-14
Ignore "no such session" in response to detach


njam:
@tomaszdurka @kris-lab @vogdb 

As in:
```
app info http-client response {"janus":"error","session_id":1790145761,"transaction":"8sicw19556","error":{"code":458,"reason":"No such session 1790145761"}}
app error Unexpected rejection error. Error: http-client error: No such session 1790145761
    at /usr/lib/node_modules/cm-janus/lib/janus/http-client.js:36:15
    at tryCatcher (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:581:18)
    at Promise._settlePromises (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:697:14)
    at Async._drainQueue (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```

PR #163, opened 2016-01-19
Reject session/plugins messages not related to "registered" plugins

tomaszdurka:


njam:
- Errors should have "code" and "reason" fields
- Process "destroyed" instead of "destroy" (anwer from janus, instead of request from browser)
- Check with Michel how the client behaves if we:
 - Respond with errors for transaction responses (e.g. "detach" - #142)
 - Respond with errors for events from janus

Should we allow any destructive operations (detach plugin, detroy session) even if there's no plugin/session?

tomaszdurka:
Discussed with @zazabe that plugin related messages should be plugin-like errors. 
We should probably intercept all session-based events with handle_id/sender and proxy or reject with normal/global error.

tomaszdurka:
Paired with @zazabe

PR #155, opened 2016-01-19
Do not treat web-socket write exceptions as errors

vogdb:
```
app error Unexpected rejection error. Error: write after end
    at writeAfterEnd (_stream_writable.js:167:12)
    at Socket.Writable.write (_stream_writable.js:214:5)
    at Socket.write (net.js:636:40)
    at Sender.frameAndSend (/usr/lib/node_modules/cm-janus/node_modules/ws/lib/Sender.js:224:22)
    at /usr/lib/node_modules/cm-janus/node_modules/ws/lib/Sender.js:126:12
    at /usr/lib/node_modules/cm-janus/node_modules/ws/lib/PerMessageDeflate.js:290:5
    at afterWrite (_stream_writable.js:361:3)
    at onwrite (_stream_writable.js:352:7)
    at WritableState.onwrite (_stream_writable.js:105:5)
    at afterTransform (_stream_transform.js:99:5)
    at TransformState.afterTransform (_stream_transform.js:74:12)
    at Zlib.callback (zlib.js:611:5)
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -62,7 +62,7 @@ Connection.prototype.isOpened = function() {
  */
 Connection.prototype._queue = function(message) {
   var self = this;
-  return new Promise(function(resolve, reject) {
+  return new Promise(function(resolve) {
     self.webSocket.once('open', function() {
       self._send(message).then(function(response) {
         resolve(response);
@@ -82,8 +82,9 @@ Connection.prototype._send = function(message) {
     this.webSocket.send(JSON.stringify(message), {}, function(err) {
       if (err) {
         reject(err);
+      } else {
+        resolve();
       }
-      resolve();
     });
   }.bind(this));
 };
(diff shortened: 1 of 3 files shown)


vogdb:
@tomaszdurka please look

tomaszdurka:
Wdyt about logging errors as debug?

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

tomaszdurka:
I think #155  and #168 are dealing with quite similar things - websocket errors. Let's try to treat them in similar way, bit let's try to keep them quite separate from whole janus topic. Means:

- Both action should imo `reject` so not `resolve` / trigger `onMessage` method
- Errors should be triggered, but handled outside of the connection (logged, etc?)
- Do not use Janus erorrs for these WebSocket errors.

So in case of #155, let's:
- Still reject as this is expected behaviour of `send`
- Catch websocket errors outside (?)

In case of #168:
- Still reject (do not call `onMessage`)
- I would emit this as an error, catch outside in different way.

vogdb:
@tomaszdurka please review

tomaszdurka:
Did you really need these timeouts? :panda_face:

vogdb:
Yes. Cause we need `message` listeners to be executed first. You can try without them :)

tomaszdurka:
Looks good.

PR #165, opened 2016-01-21
Remove stream instead of detach on hangup

tomaszdurka:


tomaszdurka:
@njam please review

njam:
@zazabe could you review this? I think you know better if this is the right thing to do!

fvovan:
lgtm

zazabe:
Nice, but currently, the video unit will detach on `hangup`, I opened an issue to support this without detaching: #2780

PR #170, opened 2016-01-21
Pass along thumbnail "createdAt" to cm

tomaszdurka:
New field "createdAt" (unix timestamp) was added to thumbnail job descriptions:
https://github.com/cargomedia/janus-gateway-rtpbroadcast/pull/57

It will be expected as a new third argument to `importVideoThumbnail` in https://github.com/cargomedia/cm/pull/2065

@tomaszdurka @vogdb can you pass it along accordingly?
Can you check if it's okay to pass superfluous arguments to the CLI call - so that we could merge this here first, and then change it in CM.

### DIFF
--- a/lib/cm-application.js
+++ b/lib/cm-application.js
@@ -15,10 +15,11 @@ function CmApplication(applicationRootPath) {
 /**
  * @param {String} channelUid
  * @param {String} thumbnailPath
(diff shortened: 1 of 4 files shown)


tomaszdurka:
@njam please review

njam:
lgtm

Issue #162, opened 2016-01-19
Extract janus client into separate codebase?


njam:
Reusable in the browser *and* in Node.js.
Let's investigate to see if and how it is feasible.

cc @tomaszdurka @zazabe

zazabe:
:+1:

vogdb:
Ok. Here is the short overview. To extract the most from cm-janus and janus.js into one codebase we need to make them work in the same way. Currently they differ. The main difference is that cm-janus is proxy and janus.js is client. I displayed this difference on the below diagram and it shows that cm-janus handles income and outcome messages independently of each other and janus.js handles them as a whole.
![cm-janus vs janus-js](https://cloud.githubusercontent.com/assets/372386/12702687/5933937c-c840-11e5-87ff-2c099ff788ef.png)
What does it mean in code:
 - cm-janus.
  ```js
  // blue connection on the diagram
  fromClientConnection.on('message', function(request) {
    if('create' == request){
      processedRequest = this.startCreateSession(request);
    }
    if('attach' == request){
      processedRequest = this.startAttachPlugin(request);
    }
    toJanusConnection.send(processedRequest);
  });

  // red connection on the diagram
  toJanusConnection.on('message', function(request) {
    if('create' == request){
      processedRequest = this.finishCreateSession(request);
    }
    if('attach' == request){
      processedRequest = this.finishAttachPlugin(request);
    }
    fromClientConnection.send(processedRequest);
  });
  ```

 - janus.js. As you can see that we wait here for responses in promises.
  ```js
   janusConnection.createSession.then(function(createdSession){})
   Session.attachPlugin.then(function(attachedPlugin){})
  ```

IMO the solution will be to change cm-janus to work as janus.js(client) cause make janus.js(client) to work as cm-janus(proxy) does not make any sense especially if we want to make this client be 3rdparty-available for open-source. It means that cm-janus should wait for response on request. If to show this on a diagram then the solution looks:
![possible solution](https://cloud.githubusercontent.com/assets/372386/12702829/6171890a-c844-11e5-8656-600eaf7ed935.png)
What does it mean in code:
 - cm-janus.
  ```js
  fromClientConnection.on('message', function(request) {
    if('create' == request){
      this.createSession(request).then(function(response){
        fromClientConnection.send(response);
      });
    }
    if('attach' == request){
      this.attachPlugin(request).then(function(response){
        fromClientConnection.send(response);
      });
    }
  });

[...]

PR #168, opened 2016-01-25
Make sure proxy doesn't crash on invalid json from websocket

vogdb:


vogdb:
@tomaszdurka please review

tomaszdurka:
Review done in #155

vogdb:
@tomaszdurka please review. I still use `JanusError.Warning` cause I don't want to close connection on invalid JSON message. Please remember that we have a single error handler which shuts connection for `error instanceof JanusError.Fatal || !(error instanceof JanusError.Error)`

tomaszdurka:
I think especially in terms of extracting client (#162) we should not use Janus errors here. I think it's nothing bad in closing connection on invalid json. It's definitely not expected behaviour from client.

tomaszdurka:
On the other hand it's not "error" in terms of cm-janus application.

What about having custom WebSocket error (or there is already?) and intercept it separately?

vogdb:
@tomaszdurka please look

PR #169, opened 2016-01-26
Remove transaction field from JanusError

vogdb:
Transaction should be only added in proxy error handler (on top of the stack). We currently are not able to pass valid transaction from methods we still want to throw `Janus` errors.

### DIFF
--- a/lib/janus/proxy.js
+++ b/lib/janus/proxy.js
@@ -61,8 +61,8 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
   var connection = proxy.createConnection(fromClientConnection, toJanusConnection);
   proxy.addConnection(connection);
 
-  var handleError = function(error, request) {
-    proxy.handleError(error, fromClientConnection, request ? request['transaction'] : null);
+  var handleError = function(error, transaction) {
+    proxy.handleError(error, fromClientConnection, transaction);
   };
 
   fromClientConnection.on('message', function(request) {
@@ -71,7 +71,7 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
         toJanusConnection.send(processedRequest);
       })
       .catch(function(error) {
-        handleError(error, request);
+        handleError(error, request ? request['transaction'] : null);
       });
   });
 
@@ -81,7 +81,7 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
         fromClientConnection.send(processedRequest);
       })
       .catch(function(error) {
-        handleError(error, request);
+        handleError(error, request ? request['transaction'] : null);
       });
   });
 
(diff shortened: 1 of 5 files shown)


vogdb:
@tomaszdurka do you mean smth like this?

tomaszdurka:
Yes, but I would remove getWebsocketMessage and only create it when we need it (proxy).

vogdb:
@tomaszdurka please look

lib/janus/proxy.js
```diff
@@ -61,20 +61,24 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
   var connection = proxy.createConnection(fromClientConnection, toJanusConnection);
   proxy.addConnection(connection);
 
-  var handleError = function(error) {
-    proxy.handleError(error, fromClientConnection);
+  var handleError = function(error, request) {
```
tomaszdurka: Make transaction arg already.

tomaszdurka:
Search and remove `message['transaction']` from Janus error constructors.

vogdb:
@tomaszdurka please look

lib/janus/proxy.js
```diff
@@ -70,7 +70,9 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
       .then(function(processedRequest) {
         toJanusConnection.send(processedRequest);
       })
-      .catch(handleError);
+      .catch(function(error) {
+        handleError(error, request ? request['transaction'] : null);
```
tomaszdurka: Looks good, but you could also pass `request['transaction']` directly.

lib/janus/proxy.js
```diff
@@ -106,15 +110,26 @@ JanusProxy.prototype.establishConnection = function(fromClientConnection, toJanu
...
-    fromClientConnection.send(error.getWebSocketMessage());
+    var errorMessage = {
+      janus: 'error',
+      error: {
+        code: error.code || 490,
+        reason: error.message
+      }
+    };
+    if (transaction) {
+      errorMessage.transaction = transaction;
```
tomaszdurka: Should we generate transaction id if not present?

lib/janus/session.js
```diff
@@ -55,7 +55,7 @@ Session.prototype.processMessage = function(message) {
 Session.prototype.onAttach = function(message) {
   var self = this;
   if (!this.pluginRegistry.isAllowedPlugin(message['plugin'])) {
-    return Promise.reject(new JanusError.IllegalPlugin(message['transaction']));
+    return Promise.reject(new JanusError.IllegalPlugin());
```
tomaszdurka: You should probably pass `message['plugin']` here.

tomaszdurka:
Rest looks good.

lib/janus/proxy.js
vogdb: no. Michel prepared janus.js to accept errors without transaction

lib/janus/session.js
vogdb: Well, `JanusError.IllegalPlugin` does not require plugin.

vogdb:
@tomaszdurka please look

vogdb:
@tomaszdurka please review

lib/janus/session.js
tomaszdurka: Ah, I looked for InvalidPlugin ...

tomaszdurka:
Looks good.

PR #173, opened 2016-02-04
Job handler: Error: stderr maxBuffer exceeded.

vogdb:
```
2016-01-22 14:35:18.235 ERROR - Error: stderr maxBuffer exceeded.
    at Socket.<anonymous> (child_process.js:806:12)
    at Socket.emit (events.js:107:17)
    at readableAddChunk (_stream_readable.js:163:16)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at Pipe.onread (net.js:540:20)
```

@tomaszdurka @vogdb will this abort the job? if so, then let's find a way to handle it more gracefully.

### DIFF
--- a/lib/job/model/abstract.js
+++ b/lib/job/model/abstract.js
@@ -1,6 +1,6 @@
 var fs = require('fs');
 var path = require('path');
-var spawn = require('child_process').spawn;
+var spawn = require('child-process-promise').spawn;
 var Promise = require('bluebird');
 var tmpName = Promise.promisify(require('tmp').tmpName);
 var rimraf = require('rimraf');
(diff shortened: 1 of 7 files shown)


vogdb:
@tomaszdurka please review

tomaszdurka:
- So there is no need to adjust job handlers as they have been using first arg only anyway?
- We should probably rename `_exec` to `_spawn` in all cases
- I am not fan of stubbing spawn the way we do:
  - Spawns actually returns event emitter, what about simply stubbing spawn with EventEmitter instance and manually trigger events on it in tests? Would simplify tests a lot.
  - Also thought about using something which has simpler interface, but is tested. What about https://www.npmjs.com/package/child-process-promise?

vogdb:
@tomaszdurka please look

tomaszdurka:
lg!

PR #179, opened 2016-02-06
Make removing of connections, sessions, plugins promise-based

vogdb:
As discussed keep `onRemove` and `_remove*` methods. Promisify them.

### DIFF
--- a/lib/janus/connection.js
+++ b/lib/janus/connection.js
@@ -1,4 +1,3 @@
-var _ = require('underscore');
 var Promise = require('bluebird');
 var Transactions = require('./transactions');
 var Session = require('./session');
@@ -91,8 +90,9 @@ JanusConnection.prototype.onCreate = function(message) {
  */
 JanusConnection.prototype.onDestroy = function(message) {
   this.transactions.add(message['transaction'], function(response) {
-    this._removeSession();
-    return Promise.resolve(response);
+    return this._removeSession().then(function() {
+      return response
+    });
   }.bind(this));
   return Promise.resolve(message);
 };
@@ -107,15 +107,18 @@ JanusConnection.prototype.onTimeout = function(message) {
 };
 
 JanusConnection.prototype.onRemove = function() {
-  this._removeSession();
-  serviceLocator.get('logger').info('Removed ' + this);
+  return this._removeSession().finally(function() {
+    serviceLocator.get('logger').info('Removed ' + this);
+  }.bind(this));
 };
 
 JanusConnection.prototype._removeSession = function() {
   if (null !== this.session) {
-    this.session.onRemove();
-    this.session = null;
+    return this.session.onRemove().finally(function() {
+      this.session = null;
+    }.bind(this));
   }
+  return Promise.resolve();
 };
 
 /**
--- a/lib/janus/connection.js
+++ b/lib/janus/connection.js
@@ -114,7 +114,7 @@ JanusConnection.prototype.onRemove = function() {
 
 JanusConnection.prototype._removeSession = function() {
   if (null !== this.session) {
-    return this.session.onRemove().finally(function() {
+    return this.session.onRemove().then(function() {
       this.session = null;
     }.bind(this));
   }
--- a/lib/janus/connection.js
+++ b/lib/janus/connection.js
@@ -107,7 +107,7 @@ JanusConnection.prototype.onTimeout = function(message) {
 };
 
 JanusConnection.prototype.onRemove = function() {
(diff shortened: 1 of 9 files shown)


vogdb:
@tomaszdurka please review

lib/janus/connection.js
```diff
@@ -107,15 +107,18 @@ JanusConnection.prototype.onTimeout = function(message) {
 };
 
 JanusConnection.prototype.onRemove = function() {
-  this._removeSession();
-  serviceLocator.get('logger').info('Removed ' + this);
+  return this._removeSession().finally(function() {
```
tomaszdurka: What is the reason to use `finally` instead of `then`? We might miss some exceptions this way.

vogdb: Well. If we did anything on exceptions, I would have used `then` but we do nothing so, I used `finally`. What shall we do on exceptions? We don't handle them anywhere.

test/spec/janus/session.js
```diff
@@ -103,16 +103,19 @@ describe('Session', function() {
   context('when removes plugin', function() {
     beforeEach(function() {
       session.plugins['plugin-id'] = sinon.createStubInstance(PluginAbstract);
+      session.plugins['plugin-id'].onRemove.returns(Promise.resolve());
```
tomaszdurka: I found instant returning resolved/rejected Promise risky :) especially with `returns(Promise.reject())` I got unhandled rejections...

vogdb: Here it should be safe

lib/janus/connection.js
tomaszdurka: Well, you cannot know from this method what will be done with the exception. Thing is now we are silently ignoring them. I prefer to have unhandled exception then to at least know about it. Agree?

vogdb:
@tomaszdurka please review

lib/janus/connection.js
```diff
@@ -107,15 +107,18 @@ JanusConnection.prototype.onTimeout = function(message) {
...
+  return this._removeSession().then(function() {
+    serviceLocator.get('logger').info('Removed ' + this);
+  }.bind(this));
 };
 
 JanusConnection.prototype._removeSession = function() {
   if (null !== this.session) {
-    this.session.onRemove();
-    this.session = null;
+    return this.session.onRemove().finally(function() {
```
tomaszdurka: Still finally. We don't want to catch/ignore exception here as well.

lib/janus/plugin/streaming.js
```diff
@@ -83,14 +83,17 @@ PluginStreaming.prototype.getStream = function() {
 
 PluginStreaming.prototype.removeStream = function() {
   if (this.stream) {
-    serviceLocator.get('streams').remove(this.getStream());
-    this.stream = null;
+    return serviceLocator.get('streams').remove(this.getStream()).finally(function() {
```
tomaszdurka: Same.

lib/janus/proxy.js
```diff
@@ -162,22 +162,21 @@ JanusProxy.prototype.addConnection = function(connection) {
  */
 JanusProxy.prototype.removeConnection = function(connection) {
   if (this.connections[connection.id]) {
-    connection.onRemove();
-    delete this.connections[connection.id];
+    return connection.onRemove().finally(function() {
```
tomaszdurka: Same.

lib/janus/session.js
```diff
@@ -86,16 +86,20 @@ Session.prototype.onDetached = function(message) {
  */
 Session.prototype._removePlugin = function(id) {
   if (this.plugins[id]) {
-    this.plugins[id].onRemove();
-    delete this.plugins[id];
+    return this.plugins[id].onRemove().finally(function() {
```
tomaszdurka: Same.

vogdb:
@tomaszdurka please review

lib/janus/session.js
```diff
@@ -86,16 +86,20 @@ Session.prototype.onDetached = function(message) {
...
 };
 
 Session.prototype.onRemove = function() {
-  _.each(this.plugins, function(plugin, id) {
-    this._removePlugin(id);
-  }.bind(this));
-  serviceLocator.get('logger').info('Removed ' + this);
+  var self = this;
+  return Promise.all(_.map(this.plugins, function(plugin, id) {
+    return self._removePlugin(id).reflect();
```
tomaszdurka: Not sure if I understand how reflect works. but I trust it.

tomaszdurka:
Looks good.

PR #180, opened 2016-02-10
Add timeout for job processing

vogdb:
Currently I'm investigating a job conversion that's running for 9 days now (!). The file is 103GB, but still this is too long.

@tomaszdurka @vogdb should we add a timeout?

cc @kris-lab

### DIFF
--- a/lib/job/model/abstract.js
+++ b/lib/job/model/abstract.js
@@ -31,6 +31,9 @@ function AbstractJob(id, jobData, configuration) {
 
   /** @type {String|null} */
   this._workingDirectory = null;
+
+  /** @type {Number} */
+  this._maxRunningTime = 24 * 60 * 60;// in seconds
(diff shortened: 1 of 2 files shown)


vogdb:
@tomaszdurka please review

njam:
Does it log an error message to the log file? Would be good to keep an eye on it!

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please be generous!

tomaszdurka:
I think this can and should stay in here - `AbstractJob` (after Manager rework too).

Let's rather reject `run` method with "Job timeout" message, rather then log it internally and cancel.

vogdb:
@tomaszdurka please look

tomaszdurka:
That looks good!

PR #189, opened 2016-03-01
Fix PluginStreaming.removeStream

vogdb:


vogdb:
@tomaszdurka please review

tomaszdurka:
Not sure about this.

I think it will fix async processor - will not have non-promise to call `.then` on, but the `removeStream` promise will never fulfil. I guess you need something more like:

```js
PluginStreaming.prototype.removeStream = function() {
  var plugin = this;
  return (new Promise(function(resolve, reject) {
    plugin.queue.push(function() {
      if (!plugin.stream) {
        return Promise.resolve();
      }
      return serviceLocator.get('streams').remove(plugin.stream);
    }, function(err) {
        plugin.stream = null;
        if (err) reject(err) else resolve();
    });
  }));
};
```

tomaszdurka:
Or to be aligned with previous version and other usages:
```js
PluginStreaming.prototype.removeStream = function() {
  var plugin = this;
  return (new Promise(function(resolve, reject) {
    plugin.queue.push(function() {
      var promise;
      if (!plugin.stream) {
        promise = Promise.resolve();
      } else {
        promise = serviceLocator.get('streams').remove(plugin.stream);
      }
      return promise.then(resolve, reject);
    });
  }));
};
```

vogdb:
@tomaszdurka please look

tomaszdurka:
lgtm

PR #197, opened 2016-03-03
Stop cm-janus process when any module fails to start

tomaszdurka:


tomaszdurka:
@vogdb please review

tomaszdurka:
Btw this kind of process exit should happen on any unhandled rejection imo. Let's not introduce it yet though as it might kill app in some cases we don't expect yet.

lib/index.js
```diff
@@ -67,14 +67,18 @@ Application.prototype.registerServices = function() {
...
   jobManager.handlerRegistry.registerFromConfiguration(this.config['jobManager']['handlersConfiguration']);
-  jobManager.start();
+
+  new Promise.all([
+    proxy.start(),
+    httpServer.start(),
+    jobManager.start()
+  ]).catch(function(error) {
+    serviceLocator.get('logger').error('Process failed. Exiting. ' + error.stack || error.message);
+    process.exit(1);
```
tomaszdurka: @vogdb should we rather emit close here?

vogdb:
ok but jobManager.start does not return Promise

vogdb:
Still your solution will work

vogdb:
lgtm

lib/index.js
vogdb: We can but this error happens on start so does not matter much. If you want to be a super correct then you can emit `close` instead.

vogdb:
lgtm

tomaszdurka:
Paired with @vogdb

PR #198, opened 2016-03-04
Allow to run cm-janus with specific modules only

tomaszdurka:
Allow to pass command-line option to run only specific modules, example usages:

```
 # runs all modules
cm-janus -c <config-file>

 # runs only server/job-manager
cm-janus -c <config-file> --server
cm-janus -c <config-file> --job-manager

# or rather
cm-janus -c <config-file> --roles=server,job-manager
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ Run services using:
 ```
 bin/cm-janus
 ```
+
+### Config
 cm-janus is based on single configuration file written in yaml format. Default config is present under [`bin/config.yaml`](bin/config.yaml).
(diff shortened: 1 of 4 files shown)


tomaszdurka:
@vogdb please review

tomaszdurka:
I went with this option:
```
cm-janus -c <config-file>  # all
cm-janus -c <config-file> --roles=server
cm-janus -c <config-file> --roles=jobs
cm-janus -c <config-file> --roles=server,jobs
```

vogdb:
lgtm

PR #201, opened 2016-03-07
Fails to start if cm can't be reached

tomaszdurka:
When booting cm virtual machine:
cc @kris-lab @njam 

```
app debug WebSocket server on port 8210 started
app debug HTTP server on port 8100 started
app info Closing application
app debug WebSocket server on port 8210 started
app debug HTTP server on port 8100 started
app info Closing application
2016-03-07 11:22:45.294 INFO - cm-api request https://example.dev.cargomedia.ch/rpc/null { method: 'CM_Janus_RpcEndpoints.removeAllStreams',
  params: [ 'mad-panda' ] }
2016-03-07 11:22:45.332 DEBUG - HTTP server on port 8100 started
2016-03-07 11:22:49.534 ERROR - Process failed. Exiting. Error: cm-api error: 500 - <h1>CM_Db_Exception</h1><h2>Database connection failed: SQLSTATE[42000] [1049] Unknown database 'sk'</h2><pre>/home/vagrant/sk/vendor/cargomedia/cm/library/CM/Db/Client.php on line 75</pre><pre>     0. {main} /home/vagrant/sk/public/index.php:0
     1. CM_Http_Handler->processRequest(CM_Http_Request_Post) /home/vagrant/sk/public/index.php:8
     2. CM_Http_Response_Abstract->factory(CM_Http_Request_Post, CM_Service_Manager) /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Http/Handler.php:21
     3. CM_Http_Response_Abstract->__construct(CM_Http_Request_Post, CM_Service_Manager) /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Http/Response/Abstract.php:311
     4. CM_Http_Request_Abstract->popPathLanguage() /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Http/Response/Abstract.php:38
     5. CM_Paging_Language_Abstract->findByAbbreviation('null') /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Http/Request/Abstract.php:201
     6. CM_Paging_Abstract->getItemsRaw() /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Paging/Language/Abstract.php:14
     7. CM_Paging_Abstract->_getItemsRaw() /home/vagrant/sk/vendor/cargomedia/cm/library/CM/Paging/Abstract.php:88
[...]

PR #200, opened 2016-03-04
Remove failed jobs working directories

tomaszdurka:
Before we decided to keep failed jobs WD to investigate, not relevant for now.

PR #199, opened 2016-03-04
Re-add failed jobs after 1 minute

tomaszdurka:

PR #193, opened 2016-03-02
Unhandled rejection

vogdb:
```
2016-03-01 09:20:12.968 ERROR - Unexpected rejection error. Error: cm-api error: 500 - Internal server error

    at /usr/lib/node_modules/cm-janus/lib/cm-api-client.js:42:13
    at tryCatcher (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:581:18)
    at Promise._settlePromises (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/promise.js:697:14)
    at Async._drainQueue (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/usr/lib/node_modules/cm-janus/node_modules/request-promise/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```
This should be intercepted rather then throwing into "global".

None of these should be cm-janus error, but simple cm-service rejection - resulting in stopping stream, etc.:
- `ERROR - Unexpected rejection error. Error: http-client error: No such session 2084716723`
- `ERROR - cm-api response { error:   { type: 'CM_Exception_NotAllowed',     msg: 'Internal server error',     isPublic: false } }`

PR #204, opened 2016-03-08
Release 0.6.3

vogdb:
- Unhandled rejection (#193, @vogdb)
- Remove failed jobs working directories (#200, @tomaszdurka)
- Re-add failed jobs after 1 minute (#199, @tomaszdurka)
- Allow to run cm-janus with specific modules only (#198, @tomaszdurka)
- Fails to start if cm can't be reached (#201, @tomaszdurka)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cm-janus",
   "description": "",
-  "version": "0.6.2",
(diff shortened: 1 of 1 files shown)


vogdb:
@tomaszdurka FYI

PR #206, opened 2016-03-09
Cannot write to buffer - blocks process from running

vogdb:
As investigated: looks like lame is exhausting node.js' STDERR buffer?

```
$ strace -p 12542
Process 12542 attached - interrupt to quit
write(2, "\33[A\33[A\33[A", 9
```

### DIFF
--- a/lib/job/model/abstract.js
+++ b/lib/job/model/abstract.js
@@ -113,7 +113,7 @@ AbstractJob.prototype._runJobScript = function(commandText) {
   serviceLocator.get('logger').debug('Starting job process: `' + commandText + '`');
   var options = {
     cwd: this.getWorkingDirectory(),
-    capture: ['stdout']
+    capture: ['stdout', 'stderr']
   };
   var self = this;
   return this._spawn(command, commandArgs, options)


vogdb:
@tomaszdurka please review

tomaszdurka:
Ok. Quick question, did you actually reproduced it? As I am not sure :)

vogdb:
Yep. Here is my playground. You can reverify.
```js
var fs = require('fs');
var path = require('path');
var spawn = require('./child-process-promise').spawn;
var Promise = require('bluebird');
var rimraf = require('rimraf');

var command;
//var commandText = 'sleep 5';
var commandText = 'echo hello';
//var commandText = 'echo foo 1>&2';
//var commandText = 'lame ./delete/sample.wav ./delete/sample.mp3';
var commandArgs = [];
if (commandText.indexOf(' ') >= 0) {
  var commandParts = commandText.split(' ');
  command = commandParts.shift();
  commandArgs = commandParts;
}

var options = {
  cwd: __dirname,
  capture: ['stdout']
};

return spawn(command, commandArgs, options)
  .progress(function(childProcess) {
/*
    console.log('[spawn] childProcess.pid: ', childProcess.pid);
    childProcess.stdout.on('data', function(data) {
      console.log('[spawn] stdout: ', data.toString());
    });
*/
    //childProcess.stderr.on('data', function(data) {
    //  console.log('[spawn] stderr: ', data.toString());
    //});
  })
  .then(function(result) {
    //console.log('done', result.stdout.toString);
    return result.stdout.toString();
  })
  .fail(function(err) {
    console.error('ERROR: ', err);
  })
  //.catch(function(err) {
  //  console.error('ERROR: ', err);
  //});
```

vogdb:
I can send to you the `sample.wav` but it breaks on simple `echo` too.

tomaszdurka:
lgtm

PR #210, opened 2016-03-17
Add fluentd to VM

njam:


njam:
@tomaszdurka @vogdb please review

Fluentd is now running in the VM.
You can then send log records to it with https://github.com/fluent/fluent-logger-node
They will show up in a file `/tmp/fluentd-output.log*`.

vogdb:
ok. Why do we configure `fluentd::config::match`?

njam:
The match's pattern is compared against the records' "tag". In this case ("**") all records will match, and they will be stored in this file `/tmp/fluentd-output.log*` - for us to debug.

Should I merge this? or you want to continue on this branch?

vogdb:
Nope. That's fine. lgtm )

PR #211, opened 2016-03-17
Add fluentd logger

vogdb:
Via https://github.com/cargomedia/puppet-cargomedia/issues/1084

See https://github.com/fluent/fluent-logger-node
Fluentd is installed in VM as of https://github.com/cargomedia/cm-janus/pull/210

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 var fs = require('fs');
 var path = require('path');
 var log4js = require('log4js');
+var FluentLogger = require('fluent-logger');
 var NestedError = require('nested-error-stacks');
 var serviceLocator = require('./service-locator');
 var CMApiClient = require('./cm-api-client');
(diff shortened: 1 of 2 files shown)


vogdb:
@tomaszdurka @njam please look. It's just a draft. Should we use an alternate way and configure fluentd to watch log4js files and get log entries from them?

njam:
I would suggest to keep "console" and "file" appenders, and make "fluentd" optional.
Maybe we can just have a log4j appender configuration in the YAML, where we can provide appenders and their values?

njam:
Regarding your question: You mean to log to file from log4j, and then configure fluentd to follow that file?
I would not do it, and rather add fluentd capabilities into cm-janus. The reason is that otherwise we add complexity in cm-janus, and we need to have a parseable log format (json?).

vogdb:
> The reason is that otherwise we add complexity in cm-janus, and we need to have a parseable log format (json?).

Ok. I think that log4js supports different layouts for its output. `json` should be supported here https://github.com/michaeltandy/log4j-json. I don't suggest to follow log4j files with fluentd. I am rather trying to understand what solution do you prefer more :)

njam:
Let's add a log4 appender for "fluentd" in cm-janus then, it's anyway very easy!

vogdb:
I agree! Should be YAML log4j configuration implemented in this issue?

vogdb:
I meant your remark
> Maybe we can just have a log4j appender configuration in the YAML, where we can provide appenders and their values?

njam:
Yes, sgtm

tomaszdurka:
Can you somehow configure log4j using file instead code? I think this would be very helpful for dev/prod separation.

njam:
Closing in favor of https://github.com/cargomedia/cm-janus/issues/213

Issue #212, opened 2016-03-17
Ability to include configuration files from directory


njam:
For easier configuration of certain parts of cm-janus with puppet it would be great if we could include multiple config files from a directory.
Similar like in [bipbip](https://github.com/cargomedia/bipbip#configuration) the config file could contain a key that specifies where to load files from:

```
include: config.d/*.yml
```

In this case cm-janus would load all matching files from that path, relativ to the current config file.

cc @tomaszdurka

njam:
Closing in favor of https://github.com/cargomedia/cm-janus/issues/213

PR #214, opened 2016-03-18
Add Config specs

vogdb:


vogdb:
To save smth that worth from #212.
@tomaszdurka please review

tomaszdurka:
lg!

PR #213, opened 2016-03-18
Log JSON

vogdb:
Each document should have:
- time (2016-03-18T22:00:00Z)
- level (error, debug, …)
- message
- any other fields (on the same level) - stacktrace, user_id, …


vogdb:
Currently `time` contains datetime in the format: `2016-03-18T14:52:24.922Z`. Should it be changed to `2016-03-18 14:52:24`?
@njam @tomaszdurka

tomaszdurka:
- can you explain a bit more what is the data-array-string magic about? (https://github.com/cargomedia/cm-janus/pull/213/files#diff-450b72d6bfe908ff2b1f73e346d0ce38R28)
- there should be message field

What should be our logger interface?
- `log('message', {extraField: 'value'})`
- `log({message: 'message', extraField: 'value'})`
or
- `log('message')` or `log({message: 'message', extraField: 'value'})` check if param is hash...

vogdb:
> - can you explain a bit more what is the data-array-string magic about? (https://github.com/cargomedia/cm-janus/pull/213/files#diff-450b72d6bfe908ff2b1f73e346d0ce38R28)

It's about [sprintf](http://www.diveintojavascript.com/projects/javascript-sprintf). Log4js supports sprtinf. That is why we are doing this magic. I just took it from https://github.com/gabe0x02/log4js-json-layout. Reworking it now.

So, what about date? You didn't answer.

vogdb:
> What should be our logger interface?

I'm afraid I didn't understand your question. We have log4js interface. https://github.com/nomiddlename/log4js-node#usage

tomaszdurka:
Yes, we use log4js interface, but as you can see log4js accepts `data` which can be either string or anything else. I think in case of simple string we should push it into `to_json`'s `message` field, and in case of object we should probably merge it to top `to_json` level. Makes sense?

vogdb:
ok

vogdb:
@tomaszdurka Will be this fine?
```js
log.info('meet me at %s', 'hello', {foo: 'bar'});
```
 => 
```json
{"time":"2016-03-18T18:01:00.799Z","level":"INFO","message":"meet me at hello","extra":{"foo":"bar"}}
```

vogdb:
@tomaszdurka please share your adjustments!

njam:
Let's use ISO 8601 datetime formatting:
```
2016-03-18T22:00:00+01:00
```
I guess it should be easy to obtain in JS!

vogdb:
Well. `toISOString` claims to use ISO 8601 but it looks different than yours.
> The toISOString() method returns a string in simplified extended ISO format (ISO 8601)
```js
var today = new Date('05 October 2011 14:48 UTC');
console.log(today.toISOString()); // Returns 2011-10-05T14:48:00.000Z
```
from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

Should I implement your format?

vogdb:
Honestly I don't insist on any method. Just want to clarify. You can easily say that you want some format and I will implement it :)

njam:
`2016-03-18T22:00:00Z` works fine too!

vogdb:
done! Please look

njam:
Ok, here's some final details about how the JSON should look:
```json
{
  "timestamp": "2016-03-23T07:40:01Z",
  "message": "my msg",
  "level": "warning",
  "tags": {
    "tag1": "v1",
    "tag2": "v2"
  },
  "extra1": "v3",
  "extra2": "v3"
}
```

Anything that we want to *segment by* should go to `tags` (all the context info!).
Additional information that's useful to understand what a record is about should be on the top level (in the example called `extra{1,2}`).

tomaszdurka:
Paired with @vogdb, @vogdb  please confirm

PR #186, opened 2016-03-16
Introduce JanusCluster service for redirection

tomaszdurka:
- Introduce JanusCluster service (http client for http://github.com/cargomedia/janus-cluster)
- Use ClusterManager service (if provided) to redirect `/edge` endpoint
- Use origin address for non-edge endpoint

PR #215, opened 2016-03-22
Add session context to most log records

vogdb:
Via https://github.com/cargomedia/puppet-cargomedia/issues/1084

@tomaszdurka @vogdb 
For debugging purposes it would be good to have a common field to identify the session of a log record.
Preferably that session context could be enriched when new information arrives. Maybe we can even parse the json "token" and extract the "web session id" to add it as well?

let's discuss!

Depends on #186 cause of websocket server url parsing


vogdb:
@tomaszdurka @njam here is a short draft.

vogdb:
@tomaszdurka @njam please compare the approaches @d68cd00 and @8057e33. What do you prefer more?

tomaszdurka:
Closing in favor of ~~#235~~ #236.

PR #217, opened 2016-03-23
Do not import empty conversion files in jobs

vogdb:


vogdb:
@tomaszdurka please look

njam:
Cross posting:
> Shouldn't we check if a stream was successfully published on CM (with rpc_publish()), because only then an archive should be imported later on?

tomaszdurka:
Thank you @vogdb for all the effort, but we don't need this.

PR #220, opened 2016-03-23
Create logger wrapper

tomaszdurka:

PR #221, opened 2016-03-24
Remove variables from log messages

vogdb:
Depends on #220

### DIFF
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -44,10 +44,10 @@ CMApiClient.prototype._request = function(action, data) {
     .then(function(response) {
       var error = response['error'];
       if (error) {
-        serviceLocator.get('logger').error('cm-api response', response);
+        serviceLocator.get('logger').error('cm-api response', {response: response});
         throw new JanusError.CmApi(error['type'] + '. ' + error['msg']);
       }
-      serviceLocator.get('logger').info('cm-api response', response.body);
+      serviceLocator.get('logger').info('cm-api response', {response: response.body});
       return response['success']['result'];
     });
 };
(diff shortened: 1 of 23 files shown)


lib/cm-api-client.js
```diff
@@ -44,10 +45,10 @@ CMApiClient.prototype._request = function(action, data) {
     .then(function(response) {
       var error = response['error'];
       if (error) {
-        serviceLocator.get('logger').error('cm-api', 'response', response);
+        serviceLocator.get('logger').error('cm-api response', response);
```
tomaszdurka: Let's wrap this into a hash.

lib/cm-api-client.js
```diff
@@ -44,10 +45,10 @@ CMApiClient.prototype._request = function(action, data) {
...
         throw new JanusError.CmApi(error['type'] + '. ' + error['msg']);
       }
-      serviceLocator.get('logger').info('cm-api', 'response', response.body);
+      serviceLocator.get('logger').info('cm-api response', response.body);
```
tomaszdurka: Same here?

lib/janus/connection.js
```diff
@@ -60,7 +60,7 @@ JanusConnection.prototype.processMessage = function(message) {
   //This is an invalid case, so lets log all such messages
   var pluginId = message['sender'] || message['handle_id'] || null;
   if (pluginId) {
-    serviceLocator.get('logger').warn('Plugin message without session: ', message);
+    serviceLocator.get('logger').warn('Plugin message without session', message);
```
tomaszdurka: Let's wrap message into `{request: message}`?

vogdb:
@tomaszdurka please review

tomaszdurka:
Fix plugin onRemove spec: `expect(logger.info.withArgs('Removed plugin', {plugin: plugin}).calledOnce).to.be.equal(true);`

tomaszdurka:
Paired with @vogdb

PR #225, opened 2016-04-01
Solve serialization depth problems

vogdb:

PR #226, opened 2016-04-03
Errors from "CMApiClient.removeStream()" are not logged

vogdb:
Update from @vogdb. Depends on #225.

We've just experienced lots of calls to `rpc_removeStream()` in CM failing ([details](https://github.com/cargomedia/cm/issues/2115)).

I noticed that there were no errors logged in cm-janus about it.

Here's an example "removeStream":
https://cargomedia.loggly.com/search/?terms=&from=2016-04-02T15%3A54%3A03.407Z&until=2016-04-02T16%3A04%3A03.407Z&source_group=

I guess the next record is the response:
```json
{"level":"INFO","message":"cm-api","additionalMessages":["response",null]}
```

@tomaszdurka @vogdb can you check if we log API call errors correctly?

### DIFF
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -39,7 +39,6 @@ CMApiClient.prototype._request = function(action, data) {
   serviceLocator.get('logger').info('cm-api request', {options: options});
   return this._requestPromise(options)
     .catch(function(reason) {
-      reason.options.body.params = reason.options.body.params.toString();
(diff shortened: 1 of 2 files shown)


vogdb:
@tomaszdurka please review.

@tomaszdurka what request to cm-api should I use to get a response HTTP.200 OK but with error field?

tomaszdurka:
lgtm

PR #229, opened 2016-04-04
Change some "info" log records to "warning" or "error"

vogdb:
Whenever something bad/unexpected happens it should be "warning" or "error".

E.g. this should be "warning":
```
{"level":"info","error":{"code":1,"message":"`ionice -c 2 -n 7 nice
```

cc @tomaszdurka

### DIFF
--- a/lib/job/manager.js
+++ b/lib/job/manager.js
@@ -56,19 +56,20 @@ JobManager.prototype.stop = function() {
  */
 JobManager.prototype._processJobFile = function(filePath) {
   var self = this;
+  var job;
   return self._readJobFile(filePath)
     .then(function(jobDescription) {
       var jobId = path.basename(filePath, '.json');
       var handler = self.handlerRegistry.get(jobDescription['plugin'], jobDescription['event']);
-      var job = handler.instantiateJob(jobId, jobDescription['data']);
+      job = handler.instantiateJob(jobId, jobDescription['data']);
       return self._processor.processUntilSuccessful(job);
     })
     .then(function() {
-      serviceLocator.get('logger').info('Removing job file', {filepath: filePath});
+      serviceLocator.get('logger').info('Removing job file', {filepath: filePath, job: job ? job.serialize() : ''});
       return fs.unlinkAsync(filePath);
     })
     .catch(function(error) {
-      serviceLocator.get('logger').error('Processing job file failed', {error: error});
+      serviceLocator.get('logger').error('Processing job file failed', {error: error, job: job ? job.serialize() : ''});
     });
 };
 
--- a/lib/job/manager.js
+++ b/lib/job/manager.js
@@ -1,3 +1,4 @@
+var _ = require('underscore');
 var path = require('path');
 var Promise = require('bluebird');
 var fs = Promise.promisifyAll(require('fs'));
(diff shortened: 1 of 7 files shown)


vogdb:
@tomaszdurka please review

lib/job/manager.js
```diff
@@ -56,19 +56,20 @@ JobManager.prototype.stop = function() {
...
     .then(function(jobDescription) {
       var jobId = path.basename(filePath, '.json');
       var handler = self.handlerRegistry.get(jobDescription['plugin'], jobDescription['event']);
-      var job = handler.instantiateJob(jobId, jobDescription['data']);
+      job = handler.instantiateJob(jobId, jobDescription['data']);
       return self._processor.processUntilSuccessful(job);
     })
     .then(function() {
-      serviceLocator.get('logger').info('Removing job file', {filepath: filePath});
+      serviceLocator.get('logger').info('Removing job file', {filepath: filePath, job: job ? job.serialize() : ''});
```
tomaszdurka: Don't pass empty string, probably better null.
Or rather `job && job.serialize()`.

tomaszdurka:
This is quite inconsistent now.
We end up with:
```js
{error: 'error', job: {job: 'job-id'}}
```

We should rather:
```js
logger.debug('foo', _.extend({error: error}, job.serialize());
```

tomaszdurka:
Method `serialize` of ours should be called differently. Something like `getContext` makes much more sense.

lib/job/manager.js
tomaszdurka: With the new approach described in https://github.com/cargomedia/cm-janus/pull/229#issuecomment-205688721 let's rather:
```js
JobManager.prototype._processJobFile = function(filePath) {
  var self = this;
  var job;
  var getJobContext = function() {
    return job ? job.serialize() : null;
  };
  return self._readJobFile(filePath)
    .then(function(jobDescription) {
      var jobId = path.basename(filePath, '.json');
      var handler = self.handlerRegistry.get(jobDescription['plugin'], jobDescription['event']);
      job = handler.instantiateJob(jobId, jobDescription['data']);
      return self._processor.processUntilSuccessful(job);
    })
    .then(function() {
      serviceLocator.get('logger').info('Removing job file', _.extend({filepath: filePath}, getJobContext()));
      return fs.unlinkAsync(filePath);
    })
    .catch(function(error) {
      serviceLocator.get('logger').error('Processing job file failed', _.extend({error: error}, getJobContext()));
    });
};
```

vogdb:
@tomaszdurka please look

lib/job/processor.js
```diff
@@ -97,7 +97,7 @@ JobProcessor.prototype._createJobWorkingDirectory = function(job) {
     fs.accessSync(dir, fs.W_OK);
   }
   catch (err) {
-    serviceLocator.get('logger').error('Job working directory could not be created');
+    serviceLocator.get('logger').error('Job working directory could not be created', job.serialize());
```
tomaszdurka: Shouldn't we add `err` too?

tomaszdurka:
Rest looks good.

vogdb:
@tomaszdurka please look

tomaszdurka:
lgtm

PR #235, opened 2016-04-05
Context for logging

tomaszdurka:
Enrich and store context on these levels:

### Proxy/httpServer
1. Connection
 - id
 - arbitary data
2. Session
 - id
3. Plugin
 - id
4. Stream
 - id
 - channelMediaId
 - channelKey

### JobManager
1. Job
 - jobId
 - handler type (e.g. `janus.plugin.cm.audioroom:archive-finished`)
 - uid

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -1,12 +1,14 @@
 var uuid = require('node-uuid');
+var Context = require('./context');
 
 /**
  * @param {String} id
  * @param {String} name
  * @param {String} data
+ * @param {Context} [context]
  * @constructor
  */
-function Channel(id, name, data) {
+function Channel(id, name, data, context) {
 
   /** @type {String} */
   this.id = id;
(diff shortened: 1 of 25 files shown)


tomaszdurka:
@vogdb please look

vogdb:
Wow. ok

vogdb:
Two questions for the start:
- Why JanusConnection has context as a field property and as an argument in the method `JanusConnection.prototype.processMessage`? It also passes it to `this.getSession(sessionId).processMessage(message, context)` but `Session.prototype.processMessage = function(message)` does not accept `context`.
- Such structures are highly dangerous `Promise.resolve([response, context])`. Better `Promise.resolve({response: response, context: context})`.

tomaszdurka:
Unfortunately I committed some discussion code - these are the questions you have.

tomaszdurka:
Closing in favor #236

PR #237, opened 2016-04-06
Make getContext return Context instead of hash

tomaszdurka:


tomaszdurka:
@vogdb please review

lib/job/manager.js
```diff
@@ -58,19 +59,21 @@ JobManager.prototype.stop = function() {
 JobManager.prototype._processJobFile = function(filePath) {
   var self = this;
   var job;
```
vogdb: `var job` is redundant now here

lib/job/model/abstract.js
```diff
@@ -133,7 +138,7 @@ AbstractJob.prototype._runJobScript = function(commandText) {
       return result.stdout.toString();
     })
     .catch(function(error) {
-      serviceLocator.get('logger').warn('Failed job process', _.extend({command: commandText}, self.getContext()));
+      serviceLocator.get('logger').warn('Failed job process', context);
```
vogdb: should we add here `error` for `context`?

lib/job/model/audioroom-recording.js
```diff
@@ -35,11 +36,11 @@ AudioroomRecordingJob.prototype._run = function() {
     .then(function(mp3File) {
       return self._audioConvert(wavFile, mp3File)
         .then(function() {
-          return serviceLocator.get('cm-application').importMediaStreamArchive(channelUid, mp3File)
+          return serviceLocator.get('cm-application').importMediaStreamArchive(channelUid, mp3File, context)
         })
         .then(function() {
           return unlink(wavFile).catch(function(error) {
-            serviceLocator.get('logger').error('Removing wave file failed', _.extend({error: error}, self.getContext()));
+            serviceLocator.get('logger').error('Removing wave file failed', context.clone().extend({error: error}));
```
vogdb: why `context.clone`?

tomaszdurka: We don't want to modify the outer context, but create one within error context/closure.

lib/job/model/abstract.js
tomaszdurka: Yep, why nots.

lib/job/manager.js
tomaszdurka: Ok.

lib/job/model/audioroom-recording.js
vogdb: but `self.getContext()` returns a new context each time

vogdb:
My main question is have you tested it locally?

vogdb:
Nonetheless changes are good to merge

lib/job/model/audioroom-recording.js
tomaszdurka: Yes we could reuse, just thought it's good practice to not.

tomaszdurka:
@vogdb Please review

vogdb:
- Why JanusHttpClient.prototype.detach and JanusHttpClient.prototype.stopStream are different in handling of context? `requestContext`
- `this.fields = _.extend(this.fields, context.fields);` => `_.extend(this.fields, context.fields);`
- `this.extend(context.fields); return this;` => `return this.extend(context.fields);`

njam:
- Add docu to README how to pass context variables

njam:
- Change parameter to: `?context={"userId":<userId>,…}` (value will be URL encoded)

njam:
- Rename "fooId" to "foo" in the context?

tomaszdurka:
Will move last comment to another issue.

tomaszdurka:
@vogdb 

- Kinda streamlined
- Ok
- Ok

vogdb:
my notes are resolved.

tomaszdurka:
@vogdb please review last commits

vogdb:
lgtm

PR #240, opened 2016-04-06
Change log file level to info

vogdb:


vogdb:
@tomaszdurka please review

PR #232, opened 2016-04-06
Reduce number of (debug) log events

vogdb:
Currently 50'000 events per hour => ~14 events per second.

screen shot 2016-04-05 at 10 18 17

Ideas:
- Combine debug records for the same transaction over some time? (instead of "-> browser", "-> janus" etc all separate?)
- What else?

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -66,7 +66,7 @@ HttpServer.prototype.installRoutes = function(router) {
           return next();
         })
         .catch(function(error) {
-          serviceLocator.get('logger').warn('Stream stop failed. Removing stream by force.', context.clone().extend({error: error}));
+          serviceLocator.get('logger').warn('Detaching plugin failed', context.clone().extend({error: error}));
           stream.plugin.removeStream()
             .then(function() {
               res.send({error: 'Stream stop failed. Stream was removed by force.'});
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -42,7 +42,7 @@ HttpServer.prototype.installRoutes = function(router) {
   var self = this;
 
   router.use(function(req, res, next) {
-    serviceLocator.get('logger').debug('request ' + req.path);
+    serviceLocator.get('logger').debug('http-server request ' + req.path);
     var serverKey = req.get('Server-Key');
     if (!serverKey || serverKey !== self.apiKey) {
       res.sendStatus(403);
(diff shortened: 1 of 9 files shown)


vogdb:
@tomaszdurka please review

tomaszdurka:
Please merge upstream. Maybe make sesne to redo the pr.

vogdb:
@tomaszdurka please review. Please pay attention to how I moved logging from `lib/connection.js` to `lib/proxy.js`

lib/http-server.js
```diff
@@ -66,8 +66,7 @@ HttpServer.prototype.installRoutes = function(router) {
           return next();
         })
         .catch(function(error) {
-          serviceLocator.get('logger').warn('Stream stop failed', context.clone().extend({error: error}));
-          serviceLocator.get('logger').info('Removing stream by force', context);
+          serviceLocator.get('logger').warn('Stream stop failed. Removing stream by force.', context.clone().extend({error: error}));
```
tomaszdurka: It's a bit messy here.

- we actually detach plugin
- we log about removing stream
- we then remove stream (normally, not by force)
- we log about removing stream by force

Ain't sure about all of this. Maybe we should rather:
- ('http-client').stopStream()
- finally plugin.removeStream()
- log about

@vogdb wdyt?

lib/janus/connection.js
```diff
@@ -61,7 +61,7 @@ JanusConnection.prototype.processMessage = function(message) {
   //This is an invalid case, so lets log all such messages
   var pluginId = message['sender'] || message['handle_id'] || null;
   if (pluginId) {
-    serviceLocator.get('logger').warn('Plugin message without session', this.getContext().extend({request: message}));
+    serviceLocator.get('logger').error('Plugin message without session', this.getContext().extend({request: message}));
```
tomaszdurka: I guess you keeping/making this as error to actually check if it ever happens? :)

tomaszdurka:
Review done

lib/janus/connection.js
vogdb: yes

lib/http-server.js
tomaszdurka: What about:

```js
serviceLocator.get('logger').warn('Detaching plugin failed', context.clone().extend({error: error}));
serviceLocator.get('logger').info('Removing stream', context);
```

tomaszdurka: Or simply:
```js
serviceLocator.get('logger').warn('Detaching plugin failed', context.clone().extend({error: error}));
```

And "Removed stream"/error will be logged from stream.js.

vogdb:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #243, opened 2016-04-07
Change Context variables: `fooId` => `foo`

vogdb:


vogdb:
@tomaszdurka please review. Especially `Channel`.

tomaszdurka:
Ok, these changes made me think about non-id approach a bit.
Before we had:
```js
{channelId: 'channel-id', channelKey: 'channel-key'}
{channelId: 'channel-id'}
{sessionId: 'session-id'}
```
With new approach we gonna have:
```js
{channel: {id: 'channel-id', key: 'channel-key'}}
{channel: {id: 'channel-id'}}
{session: 'session-id'}
```
For me above means inconsistency and it's prone to errors.

I would stick to proposition one or this:
```js
{channel: {id: 'channel-id', key: 'channel-key'}}
{channel: {id: 'channel-id'}}
{session: {id: 'session-id'}
```


@njam @vogdb wdyt?

vogdb:
> I would stick to proposition one

And what was the *proposition one* ?

tomaszdurka:
How it was before.

vogdb:
Ah, ok. I'm fine with both. On my taste this one is better

>
```js
{channel: {id: 'channel-id', key: 'channel-key'}}
{channel: {id: 'channel-id'}}
{session: {id: 'session-id'}
```

tomaszdurka:
@njam please put your cents. 

I also understand that creating those objects seems pretty basic in js/nodejs, but could be challenge in other languages? :)

njam:
- janus.sessionId
- janus.mediaId
- janus.channelKey

vogdb:
So, do you choose to keep the existing solution or partially rename? Also I think that Tomasz made a very good note:

> I also understand that creating those objects seems pretty basic in js/nodejs, but could be challenge in other languages? :)

tomaszdurka:
Done in https://github.com/cargomedia/cm-janus/pull/246

PR #246, opened 2016-04-11
Change logging schema to comply with cargomedia standard

vogdb:
See https://github.com/cargomedia/guides/tree/master/logging

### DIFF
--- a/lib/global/log4js-layout-json.js
+++ b/lib/global/log4js-layout-json.js
@@ -8,7 +8,7 @@ var isPlainObject = require('is-plain-object');
 var Log4jsJsonLayout = {
   process: function(logEvent) {
     var logEntry = {
-      time: logEvent.startTime.toISOString(),
+      timestamp: logEvent.startTime.toISOString(),
       level: logEvent.level && logEvent.level.levelStr.toLowerCase() || null
     };
 
--- a/lib/global/log4js-layout-json.js
+++ b/lib/global/log4js-layout-json.js
@@ -31,7 +31,7 @@ var Log4jsJsonLayout = {
       }
     });
     var message = this.parseMessages(messages);
-    return {message: message, janus: params};
+    return _.extend({message: message}, params);
   },
 
   parseMessages: function(messages) {
--- a/lib/global/log4js-layout-json.js
+++ b/lib/global/log4js-layout-json.js
@@ -8,7 +8,7 @@ var isPlainObject = require('is-plain-object');
 var Log4jsJsonLayout = {
   process: function(logEvent) {
     var logEntry = {
-      timestamp: logEvent.startTime.toISOString(),
+      time: logEvent.startTime.toISOString(),
       level: logEvent.level && logEvent.level.levelStr.toLowerCase() || null
     };
 
--- a/lib/global/log4js-layout-json.js
+++ b/lib/global/log4js-layout-json.js
@@ -31,7 +31,7 @@ var Log4jsJsonLayout = {
       }
     });
     var message = this.parseMessages(messages);
-    return _.extend({message: message}, params);
+    return {message: message, janus: params};
   },
 
   parseMessages: function(messages) {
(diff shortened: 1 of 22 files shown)


vogdb:
@tomaszdurka @njam  what else can you see to adjust? I'm not sure how I can do
```json
  "<app-name>": {
    "user": "<userId>",
    "extra1": "foo"
  }
```

tomaszdurka:
Let's solve #243 here. Let's nest all janus/cm-janus connected properties into `janus` namespace.
Probably we might require some changes to Context class.

vogdb:
Why `janus` and not `cm-janus` namespace? On what level do you want to nest? In Logger class or Context class. May be Log4jsJsonlayout? Possibly fluentd mappings?

tomaszdurka:
Well we discussed this with Reto and thought that:
- having namespace for app- or service-specific properties is good
- having general field for exception is good

I write above as I am still "thinking" why exception is not part of app-/service- specific fields, but guess it makes sense.

In our case:
- exception is top-level
- connection, session, plugin, jobId, etc should be within more-general "service" janus namespace

@vogdb @njam agree?

njam:
agree.

"janus" and not "cm-janus" because this is a broader concept, and these variables are valid in the scope of a few services/applications: janus-gateway, cm-janus, janus-cluster and even CM.

vogdb:
@tomaszdurka what about this method to group under 'janus' namespace?

tomaszdurka:
This is a bit more complicated. Your solution will move everything from extra into the namespace.

We won't to keep `error`, `http` and other top level fields on the top.

tomaszdurka:
Another thing. Let's not do it in JSON layout but rather in general Logger.

tomaszdurka:
Let's do it in Logger.js. Something like this?
```js

var topLevelKeys = ['timestamp', 'message', 'level', 'exception', 'httpRequest', 'computerInfo'];

Logger.prototype._log = function(level, message, context) {
  var fields = context ? context.toHash() : {};
  var context = _.pick.apply(_, [fields].concat(topLevelKeys));
  context.janus = _.omit.apply(_, [fields].concat(topLevelKeys));
  this.log4js.log(level, message, context);
};
```

tomaszdurka:
As agreed we will pass context values as they should look from top level, e.g.
```js
{janus: connectionId: 'foo'}}
```
We will use deep-merge to merge contexts.

vogdb:
@tomaszdurka please review

lib/global/log4js-layout-json.js
```diff
@@ -8,7 +8,7 @@ var isPlainObject = require('is-plain-object');
 var Log4jsJsonLayout = {
   process: function(logEvent) {
     var logEntry = {
-      time: logEvent.startTime.toISOString(),
+      timestamp: logEvent.startTime.toISOString(),
```
tomaszdurka: @njam can you confirm/explain once again about time/timestamp?

I think you said me once that timestamp is ok with fluentd, but is it?
I understand that:
- fluentd source tail for cm-janus accept time
- loggly accept timestamp

Meaning that if we pass timestamp, it will be completely ignored (treated as normal field) by fluentd, but it will be correctly intercepted by loggly.
We should rather make it complaint with fluentd and let fluentd-loggly plugin change it to timestamp, right?

Sources:
- fluentd-loggly parser producting `timestamp` field: https://github.com/patant/fluent-plugin-loggly/blob/0a47b31aa80662ff31721a93d425cd8674d70b77/lib/fluent/plugin/out_loggly_buffered.rb#L57
- fluentd cm-janus tail input: https://github.com/cargomedia/puppet-packages/blob/25d43d0826eb5edae0bb8f9f09ae38b924a53388/modules/cm_janus/manifests/init.pp#L55

tomaszdurka: Meaning we should keep `time` here, right?

njam: Sorry, should be `time`, see:
https://github.com/cargomedia/puppet-packages/blob/master/modules/cm_janus/manifests/init.pp#L55

vogdb:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #252, opened 2016-04-18
Clean up exception log levels

vogdb:
- **warning**: Problems that can appear from time to time, like communication with external services. Their overall volume should be observed to detect anomalies.
- **error**: Problems that should not usually happen. They should be investigated case-by-case.
- **fatal**: Things that should never happen, but for some reason did, and were catched with a global exception handler. These are bugs and should be fixed.

### DIFF
--- a/lib/job/model/audioroom-recording.js
+++ b/lib/job/model/audioroom-recording.js
@@ -40,7 +40,7 @@ AudioroomRecordingJob.prototype._run = function() {
         })
         .then(function() {
           return unlink(wavFile).catch(function(error) {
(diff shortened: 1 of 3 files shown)


vogdb:
@njam that is all that I could find and make lower.
@tomaszdurka please review.

vogdb:
@tomaszdurka please review

tomaszdurka:
Do they actually occur? If not I would keep them as errors, wdyt @vogdb ?

vogdb:
Well. I never saw them in the logs. It seemed to me that they fit for `warn` level. If you are fine with them as errors then it's fine for me too.

PR #262, opened 2016-05-09
Improve http server requests' logging

vogdb:
- intercept context-data sent from CM application (similar to websocket)
- change level from `debug` to `info` (https://github.com/cargomedia/cm-janus/blob/ceef69f6ff34e1db4077567a29e3acabd1676fe9/lib/http-server.js#L47)

### DIFF
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -56,10 +56,6 @@ CMApiClient.prototype._request = function(action, data, context) {
       throw new JanusError.CmApi(reason.message);
     })
     .then(function(response) {
-      if (response['context']) {
-        requestContext.extend(response['context']);
-        delete response['context'];
-      }
       requestContext.extend({janus: {response: response}});
       if (!response || (!response['error'] && !response['success'])) {
         serviceLocator.get('logger').error('cm-api unexpected response', requestContext);
--- a/lib/cm-api-client.js
+++ b/lib/cm-api-client.js
@@ -56,6 +56,10 @@ CMApiClient.prototype._request = function(action, data, context) {
       throw new JanusError.CmApi(reason.message);
     })
     .then(function(response) {
+      if (response['context']) {
+        requestContext.extend(response['context']);
+        delete response['context'];
+      }
       requestContext.extend({janus: {response: response}});
       if (!response || (!response['error'] && !response['success'])) {
         serviceLocator.get('logger').error('cm-api unexpected response', requestContext);
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -44,7 +44,7 @@ HttpServer.prototype.installRoutes = function(router) {
   var self = this;
 
   router.use(function(req, res, next) {
-    serviceLocator.get('logger').debug('http-server request ' + req.path);
+    serviceLocator.get('logger').info('http-server request ' + req.path);
     var serverKey = req.get('Server-Key');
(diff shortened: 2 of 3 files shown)


vogdb:
@tomaszdurka please review

vogdb:
> - intercept context-data sent from CM application (similar to websocket)

@tomaszdurka please tell the format of this context-data. When we receive a response from CM we take its 'success' field as the result.
```js
return response['success']['result'];
```
Under what field name we should expect CM context-data?

tomaszdurka:
Yes, we need issue for CM to send this data. Just created it https://github.com/cargomedia/cm/issues/2154.

Let's expect this data to come in `context` get parameter.  Exactly the same as in websocket server https://github.com/cargomedia/cm-janus/blob/master/lib/janus/connection.js#L166
This way we may be consistent for all types of requests (GET/POST, etc).

vogdb:
@tomaszdurka please look. Not sure about this `delete response['context'];` but otherwise we log it twice.

lib/cm-api-client.js
```diff
@@ -56,6 +56,10 @@ CMApiClient.prototype._request = function(action, data, context) {
       throw new JanusError.CmApi(reason.message);
     })
     .then(function(response) {
+      if (response['context']) {
+        requestContext.extend(response['context']);
+        delete response['context'];
```
tomaszdurka: It's actually fine to keep it on `response`. It will be just for this http response log entry (which is fine to know), while `requestContext` will go on further.

lib/cm-api-client.js
```diff
@@ -56,6 +56,10 @@ CMApiClient.prototype._request = function(action, data, context) {
       throw new JanusError.CmApi(reason.message);
     })
     .then(function(response) {
+      if (response['context']) {
```
tomaszdurka: This will be sent via get parameter. Is requestPromise merging `post` with `get`?

vogdb: I thought that we only receive context from CM. Should we send our context to it too?

lib/cm-api-client.js
vogdb: If we keep it on `response` then it will be logged twice cause we log `response` after in every log entry. Does it make sense to put it into `requestContext` twice?

lib/cm-api-client.js
tomaszdurka: Oh wait, yes. This is actually wrong place to read it. We wanted to intercept this context in the server class.

vogdb: in `lib/http-server` or `lib/janus/http-client`?

vogdb: @tomaszdurka ?

tomaszdurka: We wanted to intercept it from requests sent by CM-app, not send it to CM-app, so `lib/http-server`.

vogdb: @tomaszdurka do you mean smth like this @43f721d?

tomaszdurka: Yes. 

Also that what is done in cm-api-client is not something we should not do and we don't need. We agreed that we will send context only via the requests, not via responses. Let's remove it.

We could probably send cm-janus context within the request to cm application, but let's not do it.

vogdb:
@tomaszdurka please review

tomaszdurka:
Looks good ;)

vogdb:
Will merge and release tomorrow

PR #264, opened 2016-05-30
Include "retryCount" in job processing error

vogdb:
For alerting it would be nice if we could include the number of retries a job was tried in the log entry "Failed job process".

@vogdb wdyt, can it be easily added?

cc @kris-lab @tomaszdurka

PR #265, opened 2016-06-15
Move "retryCount" away from root level

vogdb:
Added in: https://github.com/cargomedia/cm-janus/pull/264

I would suggest to move this field away from the root level of log entries. Compare [guidelines](https://github.com/cargomedia/guides/tree/master/logging).

I'm thinking either `jobData.retryCount` or `exception.retryCount`. I would go with the second.

@tomaszdurka @vogdb wdyt?


vogdb:
I am fine with both. @tomaszdurka ?

tomaszdurka:
lgtm

PR #266, opened 2016-06-15
Update vagrant puppet provisioner to be puppet4 (environments) compatible

njam:
@vogdb please review


njam:
@cargomedia/devops fyi - there might be a few more repos where we need to update.

njam:
I don't think I broke the tests :> Are they flaky?

Anyway please merge this if you agree. Worked for me (`vagrant up`).

vogdb:
Thank you!

PR #268, opened 2016-06-15
Replace job failure's "retryCount" with "failedRepeatedly"

vogdb:
Which would be true or false.
So we can search for it in Loggly.

cc @tomaszdurka


vogdb:
@njam should we add this flat to `exception` field? or to a global scope? Currently it is `exception`.
@tomaszdurka please review

tomaszdurka:
Agree to change. 

Side-questions: was it possibly to search for `retryCount > 10`?

vogdb:
> Side-questions: was it possibly to search for `retryCount > 10`?

njam told that no.

Issue #270, opened 2016-06-21
Plugin.removeStream() cannot forcefully remove stream


njam:
Trying to remove a "corrupt" stream with the `/stopStream` API.
The following code gets executed:
```js
          stream.plugin.removeStream()
            .then(function() {
              res.send({error: 'Stream stop failed. Stream was removed by force.'});
              next();
            })
```

The API responds with "Stream stop failed. Stream was removed by force.".
But the stream seems to not be removed. I.e. it is still returned on the `/status` API.

@vogdb @tomaszdurka can you investigate?

vogdb:
ok, will look into

vogdb:
Can't reproduce. Please give a link to logs if possible.

njam:
Tomasz told me you looked at this together, and showed you the logs. He told me you have an idea what could have happened, and will implement a workaround.

vogdb:
ehhm. I don't have an idea actually :) I just found two minor problems 271 and 272 while was looking into this.

tomaszdurka:
After forced-removal:
- verify plugin's stream was removed
- verify stream was removed from streams
- otherwise log fatal error for debugging

@vogdb can you implement this in separate ticket?

kris-lab:
@tomaszdurka @vogdb what is the status here?

vogdb:
It should not happen anymore

kris-lab:
Is it closable then?

vogdb:
If @tomaszdurka does not have any objections

tomaszdurka:
Let's close it... Btw one-time problem we tried to *solve* with this never happened again.

PR #271, opened 2016-06-22
Reject with errors instead of throwing in methods that return Promises

vogdb:


vogdb:
@tomaszdurka please review. It is not the fix to #270 but it was found during it.

tomaszdurka:
lgtm

PR #272, opened 2016-06-23
drop deep-extend dependency

vogdb:
It does not extend Error objects
unclechu/node-deep-extend#26


vogdb:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #275, opened 2016-07-01
Remove all "subscribe" streams when a rtpbroadcast "publish" stream is removed

vogdb:
Goal: make sure that all "subscribes" are removed against CM's API *before* sending the "unpublish" call. This way we can avoid streamchannels with only subscribes that appear briefly after an unsubscribe.

Would it make sense to send "unsubscribe" for all subscribers in `PluginVideo.removeStream()`, if the publish stream is removed?
We could also remove the whole plugin then…

Note: the plugin is in a unusable state after unpublish anyway, because any attempts to republish will fail, because CM will find that this "mediaId" already existed.

@tomaszdurka @vogdb wdyt?

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -35,7 +35,7 @@ Channel.prototype.getContext = function() {
  * @returns {Boolean}
  */
 Channel.prototype.equalTo = function(channel) {
-  return channel && this.id == channel.id && this.name == channel.name;
+  return channel && this.id == channel.id;
 };
 
 /**
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -31,6 +31,14 @@ Channel.prototype.getContext = function() {
 };
 
 /**
+ * @param {Channel} channel
+ * @returns {Boolean}
+ */
+Channel.prototype.equalTo = function(channel) {
+  return channel && this.id == channel.id && this.name == channel.name;
+};
+
+/**
  * @param {String} name
  * @param {String} data
  * @returns {Channel}
--- a/lib/janus/plugin/video.js
+++ b/lib/janus/plugin/video.js
@@ -70,7 +70,7 @@ PluginVideo.prototype.removeStream = function() {
         promise = Promise.resolve();
       }
       if (plugin._isPublish) {
-        var subscribeStreams = serviceLocator.get('streams').findByChannel(plugin.stream.channel);
+        var subscribeStreams = serviceLocator.get('streams').findAllByChannel(plugin.stream.channel);
         subscribeStreams = _.without(subscribeStreams, plugin.stream);
         promise = Promise.map(subscribeStreams, function(subscribeStream) {
           return subscribeStream.plugin.removeStream();
--- a/lib/janus/plugin/video.js
+++ b/lib/janus/plugin/video.js
@@ -73,9 +73,9 @@ PluginVideo.prototype.removeStream = function() {
         var subscribeStreams = serviceLocator.get('streams').findByChannel(plugin.stream.channel);
         subscribeStreams = _.without(subscribeStreams, plugin.stream);
         promise = Promise.map(subscribeStreams, function(subscribeStream) {
-          return subscribeStream.plugin.removeStream().then(function() {
-            return serviceLocator.get('http-client').detach(subscribeStream.plugin);
-          });
+          return subscribeStream.plugin.removeStream();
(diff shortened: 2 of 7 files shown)


vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please hold

vogdb:
> As pub/sub concept doesn't really fit into Audio plugin.

@tomaszdurka I am a bit confused. I am looking now into the code of Audio plugin and it uses the concept of pub/sub. That is why I did the halt of review. Please look at `PluginAudio.prototype.onChangeroom`. It does `subscribe`. Is it a mistake there?

njam:
In CM streamchannels always can have *subscribe streams* and *publish streams*. For the audioroom plugin we decided to use only *subscribe streams*.
(On the other hand for the rtpbroadcast plugin the sending party is a *publish stream* and the watching parties are *subscribe streams*.)

vogdb:
Ok. When we remove publisher, we remove its subscribers first. It means that we want to remove also subscribers' audio streams but not publisher's audio stream. It means that we need to differ between publisher's and subscriber's audio. How can I solve this problem without marking audio streams as publisher/subscriber too?

njam:
No, for audioroom there is always *only subscribers*.
The logic in this ticket should only apply to rtpbroadcast.

vogdb:
@tomaszdurka please review

lib/janus/plugin/video.js
```diff
@@ -45,6 +48,44 @@ PluginVideo.prototype.processMessage = function(message) {
...
+      var promise;
+      if (!plugin.stream || !plugin._isPublish) {
+        promise = Promise.resolve();
+      }
+      if (plugin._isPublish) {
+        var subscribeStreams = serviceLocator.get('streams').findByChannel(plugin.stream.channel);
+        subscribeStreams = _.without(subscribeStreams, plugin.stream);
+        promise = Promise.map(subscribeStreams, function(subscribeStream) {
+          return subscribeStream.plugin.removeStream().then(function() {
+            return serviceLocator.get('http-client').detach(subscribeStream.plugin);
```
tomaszdurka: We should probably do not detach plugin here. That should be done automatically via janus, we should probably not duplicate this behaviour here. If not probably it should be done in `removeStream` for all.

lib/janus/plugin/video.js
```diff
@@ -45,6 +48,44 @@ PluginVideo.prototype.processMessage = function(message) {
 };
 
 /**
+ * @inheritDoc
+ */
+PluginVideo.prototype.publish = function() {
+  return PluginVideo.super_.prototype.publish.call(this)
+    .then(function(result) {
+      this._isPublish = true;
```
tomaszdurka: We should probably set this back to null/false when removing stream. Plus set it to false when subscribing?

lib/channel.js
```diff
@@ -31,6 +31,14 @@ Channel.prototype.getContext = function() {
 };
 
 /**
+ * @param {Channel} channel
+ * @returns {Boolean}
+ */
+Channel.prototype.equalTo = function(channel) {
+  return channel && this.id == channel.id && this.name == channel.name;
```
tomaszdurka: This equal to is a bit dangerous. Maybe not equal but possibility to store channels with different id, but the same names, etc. Not sure if we should care.

vogdb: What are the alternatives to this? Shall I short this only to `return channel && this.id == channel.id`?

lib/janus/plugin/video.js
vogdb: It is `false` by default, so no reason to do it again in `subscribe`.

tomaszdurka: Agree, but let's put it back to false when `removeStream`

lib/channel.js
vogdb: @tomaszdurka ???

vogdb:
@tomaszdurka all your notes are fixed except the one about channel equality. Please answer my question there.

tomaszdurka:
Let's use media-id only. This is the one which links together streams in cm-janus.

Additional note: rename findByChannel to findAllByChannel to streamline to our conventions :)

tomaszdurka:
lgtm

tomaszdurka:
Let's merge this, tag/release it quickly.

PR #278, opened 2016-07-12
Rework stopStream

vogdb:


vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

vogdb:
@tomaszdurka please review

lib/http-server.js
```diff
@@ -72,16 +72,24 @@ HttpServer.prototype.installRoutes = function(router) {
           serviceLocator.get('logger').warn('Detaching plugin failed', context.clone().extend({exception: error}));
           stream.plugin.removeStream()
             .then(function() {
-              res.send({error: 'Stream stop failed. Stream was removed by force.'});
-              next();
+              if (stream.plugin.stream || serviceLocator.get('streams').has(streamId)) {
```
tomaszdurka: Shall we try to remove stream from streams if it's there only. I somehow think it was the problem once. That we didn't have stream on plugin, but we had it in the streams. Here we could force removal from the streams.

vogdb: if it hasn't been removed by the time of that call, it is an error and there is a very high possibility that the second call to `streams.remove` will just cause this error the second time. Are you sure you want to do it?

tomaszdurka: Not sure. This was quick thought. Earlier I thought we could somehow improve removal of the stream itself. Maybe if stream is not removed (from either plugin or streams) we should detach plugin? I am worried about this once incident we had with infinite removal of the stream.

vogdb: But we do this removal of stream here cause the detachment of `stream.plugin` failed. It would be odd to retry either detach of plugin or remove of stream.

vogdb: @tomaszdurka ?

tomaszdurka: Not sure about this one. We can merge it, not sure if it will help though. Let's merge it.

vogdb:
@tomaszdurka FYI

PR #280, opened 2016-07-15
AbstractJob context.janus is overwritten by child classes

vogdb:
Reason:
image

Seems that our `var deepExtend = require('util')._extend;` doesn't work as expected.
https://github.com/cargomedia/cm-janus/blob/master/lib/context.js#L2

Did experiment:
```
tomasz@panda:~$ node
> var deepExtend = require('util')._extend;
undefined
> var hash = {janus: {jobId: 1}};
undefined
> deepExtend(hash, {janus: {jobData: {foo: 'bar'}}});
{ janus: { jobData: { foo: 'bar' } } }
> console.log(hash);
{ janus: { jobData: { foo: 'bar' } } }
undefined
```

Maybe let's go back to `deepExtend` after all?

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -1,5 +1,5 @@
 var _ = require('underscore');
-var deepExtend = require('util')._extend;
+var deepExtend = require('deep-extend');
 
 function Context(fields) {
   this.fields = _.clone(fields) || {};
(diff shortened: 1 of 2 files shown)


vogdb:
@tomaszdurka please review. We are using the custom version from my repo. It contains the patch unclechu/node-deep-extend#26 that can not be merged.

tomaszdurka:
lgtm

tomaszdurka:
Shall we add test for this case?

PR #285, opened 2016-08-01
Use International System of Units for config values

vogdb:


vogdb:
@tomaszdurka please review

tomaszdurka:
Do you think it's better? I thought JS is expecting miliseconds.

tomaszdurka:
Please decide, looks good.

PR #286, opened 2016-08-09
Do not send stdout/stderr on failedRepeatedly exceptions

vogdb:
just to be sure, we log every time, but we add stdout/stdeer every Xth time? or we log nothing or everything every Xth time?

I would log nothing and everything every Xth time

### DIFF
--- a/lib/job/model/abstract.js
+++ b/lib/job/model/abstract.js
@@ -143,11 +143,11 @@ AbstractJob.prototype._runJobScript = function(commandText) {
     })
     .catch(function(error) {
       self._retryCount++;
-      if (self._retryCount > 10) {
+      if (self._retryCount % 10 == 0) {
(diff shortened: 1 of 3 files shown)


vogdb:
@tomaszdurka please review

tomaszdurka:
Looks good!

njam:
I understand the business logic is now:
- Only log every 10th trial (0, 10, 20, …)
- Whenever something is logged then `failedRepeatedly` is set

Did I interpret that correct?
If so, then I don't think this is what we want.

I would suggest:
- Log every incident
- Add `failedRepeatedly` at/after the 10th trial
- Add command output only the first trial

vogdb:
> Did I interpret that correct?

How it works now
- log every incident as a warning without command output
- Log incident as an error with command output every 10x times.

